PR #40, opened 2019-07-12
Analytics CI

LoekL:
This PR contains all the CI related things for the Analytics Pipeline.

Buildkite pipeline currently failing as it's missing the Docker files it needs (will be part of a separate PR).


ci/analytics-endpoint.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ "${STATUS_CODE}" != "200" ]; then echo 'Error: v1/file did not return 200!' && exit 1; fi;
+
+finish() {
+  # Stops and removes all containers:
+  docker-compose -f services/docker/docker_compose_local_analytics.yml down
+  docker-compose -f services/docker/docker_compose_local_analytics.yml rm --force
+  # Remove /tmp/ci-online-services:
+  rm -rf /tmp/ci-online-services || exit 0
+}
+trap finish EXIT
```
benclive: I think this might need to be higher up? trap effectively registers the callback so you probably need to register early in order to trap the failures

LoekL: _NOTE: It is important that the trap statement be placed at the beginning of the script to function properly. Any commands above the trap can exit and not be caught in the trap._

Source: https://www.putorius.net/using-trap-to-exit-bash-scripts-cleanly.html

Moved up!

PR #43, opened 2019-07-12
Analytics python

LoekL:
This PR contains all Python code:

- To run the Endpoint.
- To run the Cloud Function.
- To run backfills for the Cloud Function (dataflow).

Note that Buildkite won't pass _until_ the Docker / K8s PR is merged into master.


services/python/analytics-pipeline/src/endpoint/common/functions.py
```diff
@@ -0,0 +1,27 @@
...
+        try:
+            event_new['eventAttributes'] = json.dumps(event['eventAttributes'])
+        except Exception:
+            pass
+        return event_new
+    except Exception:
+        return event
+
+
+def get_date_time():
```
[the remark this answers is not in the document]
LoekL: We only never used `ts`, so I removed it.

I do want all of those things to be captured at exactly the same point in time, not across different function calls, so I'll leave as-is otherwise!

services/python/analytics-pipeline/src/endpoint/common/functions.py
```diff
@@ -0,0 +1,27 @@
+import datetime
+import json
+import time
+
+
+def format_event(index, event, batch_id, analytics_environment):
```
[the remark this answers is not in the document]
LoekL: I changed it to return a bool with the event, and changed all related code in `main.py` accordingly! Renamed it to `try_format_event` to better reflect that the operation can fail.

services/python/analytics-pipeline/src/function/main.py
```diff
@@ -0,0 +1,126 @@
...
+    new_list = [
+      {'job_name': job_name,
+       'processed_timestamp': time.time(),
+       'batch_id': hashlib.md5('/'.join(gspath.split('/')[-2:]).encode('utf-8')).hexdigest(),
+       'analytics_environment': path_parser(gspath, 'analytics_environment='),
+       'event_category': path_parser(gspath, 'event_category='),
+       'event_ds': path_parser(gspath, 'event_ds='),
+       'event_time': path_parser(gspath, 'event_time='),
+       'event': element_cast(i),
+       'file_path': gspath} for i in list]
```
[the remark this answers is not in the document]
LoekL: Changed to `for event in events_list`.

services/python/analytics-pipeline/src/function/main.py
```diff
@@ -0,0 +1,126 @@
...
+import os
+
+from common.functions import json_parser, parse_field, path_parser, unix_timestamp_check
+from common.bigquery import provision_bigquery
+
+# Function acts as function-gcs-to-bq@[your project id].iam.gserviceaccount.com
+client_gcs, client_bq = storage.Client(), bigquery.Client()
+
+
+def element_cast(element):
```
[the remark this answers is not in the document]
LoekL: I believe this is now updated to:

```
def cast_to_string(element):
    if isinstance(element, (list, dict)):
        return json.dumps(element)
    else:
        return str(element)
```

services/python/analytics-pipeline/src/endpoint/common/functions.py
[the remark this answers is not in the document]
LoekL: I added a description with this context as well.

services/python/analytics-pipeline/src/function/main.py
```diff
@@ -0,0 +1,126 @@
...
+    errors = client_bq.insert_rows(table_logs, event_formatter(['parse_initiated'], os.environ['FUNCTION_NAME'], gspath))
+    if errors:
+        print('Errors while inserting logs: ' + str(errors))
+
+    # Get file from GCS:
+    bucket = client_gcs.get_bucket(bucket_name)
+    batch = json_parser(bucket.get_blob(file_name).download_as_string().decode('utf8'))
+
+    # If dict nest in list:
+    if isinstance(batch, dict):
```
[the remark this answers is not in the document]
LoekL: Did that, now it's using the `[boolean, result]` tactic we're also using in the endpoint's `try_format_event` function. Accordingly this function is renamed to `try_parse_json`.

services/python/analytics-pipeline/src/function/main.py
```diff
@@ -0,0 +1,126 @@
...
+        batch = [batch]
+
+    # Parse list:
+    if isinstance(batch, list):
```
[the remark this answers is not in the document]
LoekL: Removed it, as `try_parse_json` now always returns a list.

services/python/analytics-pipeline/src/function/main.py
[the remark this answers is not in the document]
LoekL: Changed to `for event in events_list`.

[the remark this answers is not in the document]
LoekL: Changed to `for event in event_list`.

services/python/analytics-pipeline/src/endpoint/main.py
```diff
@@ -0,0 +1,122 @@
...
+        'message': 'Exception: {}'.format(e)})
+    response.status_code = http_client.INTERNAL_SERVER_ERROR
+    return response
+
+
+if __name__ == '__main__':
+    # This is triggered when running locally (e.g. `python main.py`).
+    # Gunicorn is used to run the application on Google App Engine / Kubernetes (Container),
+    # and separately configured to handle threading/parallel/async requests -> see entrypoint in Dockerfile.
+    app.run(host='localhost', port=8080, debug=True, threaded=True)
```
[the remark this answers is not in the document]
LoekL: What do you mean exactly?

services/python/analytics-pipeline/src/function/main.py
```diff
@@ -0,0 +1,126 @@
...
+import hashlib
+import base64
+import json
+import time
+import os
+
+from common.functions import json_parser, parse_field, path_parser, unix_timestamp_check
+from common.bigquery import provision_bigquery
+
+# Function acts as function-gcs-to-bq@[your project id].iam.gserviceaccount.com
```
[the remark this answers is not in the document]
LoekL: Changed to `# Function acts as the service account named function-gcs-to-bq@[your project id].iam.gserviceaccount.com`

services/python/analytics-pipeline/src/endpoint/main.py
```diff
@@ -0,0 +1,122 @@
...
+client_storage = storage.Client.from_service_account_json(os.environ['SECRET_WRITER_JSON'])
+bucket = client_storage.get_bucket(os.environ['BUCKET_NAME'])
+
+try:
+    try:
+        subprocess.check_call('base64 --decode %s > /tmp/analytics-gcs-writer.p12' % os.environ['SECRET_WRITER_P12'], shell=True)
+    except Exception:
+        subprocess.call('cp %s /tmp/analytics-gcs-writer.p12' % os.environ['SECRET_WRITER_P12'], shell=True)
+
+    subprocess.call('openssl pkcs12 -passin pass:notasecret -in /tmp/analytics-gcs-writer.p12 -nodes -nocerts > /tmp/analytics-gcs-writer.pem', shell=True)
```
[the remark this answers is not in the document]
LoekL: This part runs only once, when you boot up the service. It's to convert a .p12 key into DER format which I need to sign URLs (the endpoint's `v1/file` method. I do not know where else I could/would do this conversion..

Flow: I pull a container from k8s, mount the .p12 file in it & use `main.py` as the ENTRYPOINT.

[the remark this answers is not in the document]
LoekL: FYI - You cannot generate service account keys in DER format with GCP (not in their UI, not with Terraform, etc.). So what we _could_ do is tell users to download a .p12 SA key, run a script to convert it, and finally a command to upload this key as a secret in k8s. This vs. the current setup where they don't have to do any of this, but at the cost of these system calls..

.gitignore
```diff
@@ -5,3 +5,7 @@
 **/.idea/
 **/*.csproj.user
 **/*.DotSettings.user
+*.DS_Store
```
JoelOtter: You don't need the `*` here; in fact the latest `.gitignore` already ignores these files so you can just rebase and delete this line,

services/python/analytics-pipeline/src/requirements/scale-test.txt
```diff
@@ -0,0 +1,2 @@
+# Scale Test
+google-cloud-storage==1.14.0
```
JoelOtter: You're using a different version of the `google-cloud-storage` package here?

.gitignore
```diff
@@ -5,3 +5,7 @@
...
+**/venv*/
+**/.python-version
```
JoelOtter: Why is this in the `gitignore` and also in the file tree?

services/python/analytics-pipeline/.gitignore
```diff
@@ -0,0 +1,4 @@
+*.zip
```
JoelOtter: generally speaking I'd prefer to have this stuff in the root `gitignore`, with leading paths for specificity if needed. Easier to find.

.gitignore
```diff
@@ -5,3 +5,7 @@
 **/.idea/
 **/*.csproj.user
 **/*.DotSettings.user
+*.DS_Store
+**/venv*/
```
JoelOtter: Doesn't one usually put virtualenvs out-of-tree? E.g. in `~/.virtualenvs/`.

If we're going to leave this here I'd rather it be specific to a directory.

.gitignore
```diff
@@ -5,3 +5,7 @@
...
+**/.python-version
+**/vault.sh
```
JoelOtter: I don't understand what this is for.

services/python/analytics-pipeline/.python-version
```diff
@@ -0,0 +1 @@
+3.7.1
```
JoelOtter: Mentioned above but: confused why this file is here when it's also gitignored

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
JoelOtter: you have `debug=True` in here

services/python/analytics-pipeline/src/dataflow/setup.py
```diff
@@ -0,0 +1,80 @@
...
+        for command in CUSTOM_COMMANDS:
+            self.RunCustomCommand(command)
+
+
+# apache-beam is by default installed on dataflow workers
+REQUIRED_PACKAGES = []
+
+setuptools.setup(
+    name='common',
+    version='0.0.1',
```
JoelOtter: Our version for online services is currently `0.1.0`, should we align this with that? I may have misunderstood what this means. 

(We may want to bump the version number when analytics ships; worth asking Nathan maybe?)

LoekL: Aligned to `0.1.0` for now & I've asked Nathan whether we are intending to bump it when we merge analytics in. So pending his response might change it to something higher than `0.1.0`. Will then also sync with you to update any other file(s) of the Gateway & Deployment Pool.

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
LoekL: Ah, somehow I did not capture that :/

Following: https://www.tutorialspoint.com/flask/flask_application

_A Flask application is started by calling the run() method. However, while the application is under development, it should be restarted manually for each change in the code. To avoid this inconvenience, enable debug support. The server will then reload itself if the code changes. It will also provide a useful debugger to track the errors if any, in the application._

I'll therefore just remove it.

services/python/analytics-pipeline/src/dataflow/setup.py
JoelOtter: Nice one, cheers. Would suggest `0.2.0` as this represents a large feature addition. :)

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
JoelOtter: Cool, yeah, no point having this in a container

services/python/analytics-pipeline/src/function/main.py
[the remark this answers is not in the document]
JoelOtter: Can we use `your Google project id` instead, here and throughout? We don't want customers confusing this with their SpatialOS project.

.gitignore
LoekL: Done

services/python/analytics-pipeline/src/requirements/scale-test.txt
LoekL: Good catch, that wasn't necessary. Updated to `1.16.1` like the others.

services/python/analytics-pipeline/.gitignore
LoekL: Moved everything to root `.gitignore`.

services/python/analytics-pipeline/.python-version
LoekL: Shouldn't be there, deleted. Probably happens when I committed it before, and only afterwards added to .gitignore? Although you'd expect git would then see it as a deleted file..

services/python/analytics-pipeline/src/dataflow/setup.py
LoekL: His preliminary response is _"if we add major functionality we’d go to 0.2.0 and the 0.1.x would be for iterative releases or bug fixes/updates"_, so that corresponds with your proposal. 

He's going to check with Eric to see if we have a company policy around versioning before making a final call

.gitignore
LoekL: Shouldn't have been the case - they should all be gone now.

.gitignore
LoekL: It was just a file I had for mounting the required CI secrets into Vault. I've placed the file outside of the OS repo & removed the reference.

LoekL: Do we want to somehow put this in one single place for the entire team? Potentially including the actual secret key files used?

.gitignore
LoekL: They are in `/docs/analytics-pipeline/` because I wrote the original docs as if the user was in this particular directory.

As the docs will be overhauled, this might no longer be the case. I also do not have a strong opinion on taking them out-of-tree. 

Wasn't aware they generally went into `~/.virtualenvs/`, in Google's docs they generally just create it in whichever directory the user currently is, for example they suggest: 

```bash
python2.7 -m virtualenv env
source env/bin/activate
```

Taken from: https://cloud.google.com/dataflow/docs/quickstarts/quickstart-python

For now I've made them specific to a directory: `docs/analytics-pipeline/venv-*`

.gitignore
LoekL: For instance, the `ci-vault.sh` just contained:

```bash
# Hashicorp Vault // Secrets Management

# https://github.com/improbable/platform/blob/master/docs/product-groups/dev-workflow/buildkite/how-to-use-secrets-in-ci.md
# Create required secrets in Vault:

imp-vault store --product_group="dev-workflow" --environment="production" --role="buildkite" \
  --in_use_by="buildkite-agents" --type_name="gce-key-pair" --name="ci/improbable/logical-flame-194710/analytics-gcs-writer-json" \
  --path="/Users/loek/secrets/logical-flame-194710/analytics-gcs-writer.json"

imp-vault store --product_group="dev-workflow" --environment="production" --role="buildkite" \
  --in_use_by="buildkite-agents" --type_name="generic-token" --name="ci/improbable/logical-flame-194710/analytics-gcs-writer-p12" \
  --path="/Users/loek/secrets/logical-flame-194710/analytics-gcs-writer-b64.p12"

imp-vault store --product_group="dev-workflow" --environment="production" --role="buildkite" \
  --in_use_by="buildkite-agents" --type_name="gce-key-pair" --name="ci/improbable/logical-flame-194710/analytics-endpoint-json" \
  --path="/Users/loek/secrets/logical-flame-194710/analytics-endpoint.json"

imp-vault store --product_group="dev-workflow" --environment="production" --role="buildkite" \
  --in_use_by="buildkite-agents" --type_name="generic-token" --name="ci/improbable/logical-flame-194710/api-key" \
  --path="/Users/loek/secrets/logical-flame-194710/api-key.json"

# Templates
# - generic-token: {"token":"change-me","ownerEmail":"change-me","ownerDetails":"change-me","description":"change-me"}

# Note - first base64 encode the p12 key before adding it as the value of "token":
base64 /Users/loek/secrets/logical-flame-194710/analytics-gcs-writer.p12 > /Users/loek/secrets/logical-flame-194710/analytics-gcs-writer-b64.p12
```

services/python/analytics-pipeline/src/function/main.py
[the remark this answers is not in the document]
LoekL: Yes!

services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -0,0 +1,158 @@
...
+    gs://[your Google project id]-analytics/data_type=json/analytics_environment=function/...
+
+    When for instance passing the above path & 'data_type=' as the key it will return its value 'json'.
+    """
+
+    try:
+        # Try to split the path by key (indexing to 1 will fail if key not present in path):
+        value = path.split(key)[1].split('/')[0]
+        if key == 'event_ds=' and not validate_date(value):
+            value = None
```
JoelOtter: flow here is weird. just `return None` here and L46

services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -0,0 +1,158 @@
...
+    except Exception:
+        value = None
+    return value
+
+
+def parse_analytics_environment(environment):
+    if environment in ['all', '']:
```
JoelOtter: What happens if this function receives `None`? Can that happen?

services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -0,0 +1,158 @@
...
+        raise Exception("event-time argument must be one of: {'all', '0-8', '8-16', '16-24'}")
+
+    if time_part == 'all':
+        time_part_list, name_time = ['0-8', '8-16', '16-24'], '{time_part}-times'.format(time_part=time_part)
+    else:
+        time_part_list, name_time = [time_part], time_part
+    return time_part_list, name_time
+
+
+def convert_list_to_sql_tuple(sql_filter_list):
```
JoelOtter: will this work with nested lists? does it need to?

services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -0,0 +1,158 @@
...
+    if isinstance(sql_filter_list, list):
+        return str(sql_filter_list).replace('[', '(').replace(']', ')')
+    else:
+        raise Exception('listToSqlTuple did not receive a list!')
```
JoelOtter: `TypeError`?

services/python/analytics-pipeline/src/dataflow/common/bigquery.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    from common.bigquery_schema import bigquery_table_schema_dict
+    from google.cloud.exceptions import NotFound
+    from google.cloud import bigquery
+
+    def asset_exists(client, asset_type, reference):
+
+        if asset_type not in ['dataset', 'table']:
+            raise Exception('Error - asset_type must be one of: {dataset, table}!')
```
JoelOtter: `TypeError`?

services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def try_parse_json(text):
```
JoelOtter: I feel like this return signature is kinda weird. Wouldn't it be better to simply return the result, _or_ throw an exception?

services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -0,0 +1,158 @@
...
+                result = [result]
+            return [True, result]
+
+        # Otherwise, fail:
+        except Exception as e:
+            result = 'Error: {e} -- The following string could not be parsed as JSON: {text}'.format(e=e, text=text)
+            return [False, [result]]
+
+
+def parse_dict_key(event_dict, option1, option2=''):
```
JoelOtter: Document this, I don't understand it

services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # If timestamp is in human readable string format, try to parse using the given
+    # formats, extract the unix timestamp if the timestamp is valid & return it:
+    timestamp_list = []
+    if isinstance(timestamp, str):
+        for format in timestamp_format_list:
+            try:
+                d = datetime.datetime.strptime(timestamp, format)
+                timestamp_list.append(time.mktime(d.timetuple()))
+            except Exception:
```
JoelOtter: I reckon use the specific error type that `strptime` throws, if you can.

services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -0,0 +1,158 @@
...
+        return timestamp
+
+    # If timestamp is in human readable string format, try to parse using the given
+    # formats, extract the unix timestamp if the timestamp is valid & return it:
+    timestamp_list = []
+    if isinstance(timestamp, str):
+        for format in timestamp_format_list:
+            try:
+                d = datetime.datetime.strptime(timestamp, format)
+                timestamp_list.append(time.mktime(d.timetuple()))
```
JoelOtter: surely if the conversion works you just want to return here? and then L151 could be `return None` or, better, `raise Exception` ?

services/python/analytics-pipeline/src/dataflow/p1_gcs_to_bq_backfill.py
```diff
@@ -0,0 +1,148 @@
...
+parser.add_argument('--event-ds-start', dest='event_ds_start', default='2019-01-01')
+parser.add_argument('--event-ds-stop', dest='event_ds_stop', default='2020-12-31')
+parser.add_argument('--event-time', dest='event_time', default='all')  # {0-8|8-16|16-24}
+parser.add_argument('--scale-test-name', dest='scale_test_name', default='')
+
+args = parser.parse_args()
+
+if args.event_ds_start > args.event_ds_stop:
+    print('Error: ds_start cannot be later than ds_stop!')
+    sys.exit()
```
JoelOtter: Exit with a code, in case someone uses this as part of a script

services/python/analytics-pipeline/src/dataflow/p1_gcs_to_bq_backfill.py
```diff
@@ -0,0 +1,148 @@
...
+      'job_name': job_name
+      })
+    pipeline_options.view_as(SetupOptions).save_main_session = True
+
+    p1 = beam.Pipeline(options=pipeline_options)
+    fileListGcs = (p1 | 'CreateGcsIterators' >> beam.Create(list(generate_gcs_file_list(generate_date_range, args.event_ds_start, args.event_ds_stop, args.gcs_bucket, environment_list, args.event_category, time_part_list, args.scale_test_name)))
+                      | 'GetGcsFileList' >> beam.ParDo(GetGcsFileList())
+                      | 'GcsListPairWithOne' >> beam.Map(lambda x: (x, 1)))
+
+    # fileListGcs | 'dumpGCSFileList' >> beam.io.WriteToText('gs://{gcs_bucket}/data_type=dataflow/batch/output/{job_name}/0_fileListGcs'.format(gcs_bucket = args.gcs_bucket, job_name = job_name)) # Cloud-Debug [when using DataflowRunner]
```
JoelOtter: remove all these commented-out lines?

services/python/analytics-pipeline/src/function/main.py
```diff
@@ -0,0 +1,111 @@
+# -*- coding: utf-8 -*-
```
JoelOtter: Why do some files have this preamble and some do not? Are they necessary?

services/python/analytics-pipeline/src/dataflow/setup.py
LoekL: `0.2.0` it is -> https://docs.improbable.io/reference/14.0/shared/release-policy#versioning

.gitignore
```diff
@@ -5,4 +5,12 @@
 **/.idea/
 **/*.csproj.user
 **/*.DotSettings.user
-.DS_Store
\ No newline at end of file
+**/__pycache__
+**/common.egg-info
+*.zip
+docs/analytics-pipeline/venv-*
```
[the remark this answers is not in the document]
LoekL: Because I wrote the docs in the context of being in `docs/analytics-pipeline/`. So if I'd at that point execute a `python -m virtualenv venv-endpoint` it would be created in that directory..

I guess if the docs are no longer going to be hosted in an easy/logical way to consume in GitHub but on the website instead, I'll just remove the line for now. We can see what we do when we take care of the docs.

.gitignore
```diff
@@ -5,4 +5,12 @@
 **/.idea/
 **/*.csproj.user
 **/*.DotSettings.user
-.DS_Store
\ No newline at end of file
+**/__pycache__
+**/common.egg-info
+*.zip
```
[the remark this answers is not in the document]
LoekL: Changed to `services/python/analytics-pipeline/cloud-function-analytics.zip`

services/python/analytics-pipeline/src/dataflow/common/bigquery.py
```diff
@@ -0,0 +1,119 @@
+
```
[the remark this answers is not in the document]
LoekL: Removed

services/python/analytics-pipeline/src/dataflow/common/bigquery_schema.py
```diff
@@ -0,0 +1,32 @@
...
+  bigquery.SchemaField(name='batch_id', field_type='STRING', mode='NULLABLE', description='MD5 hash of the GCS filepath.'),
+  bigquery.SchemaField(name='event_id', field_type='STRING', mode='NULLABLE', description='MD5 hash of the GCS filepath + "/{event_index_in_batch}".'),
+  bigquery.SchemaField(name='event_index', field_type='INTEGER', mode='NULLABLE', description='The index of the event within its batch.'),
+  bigquery.SchemaField(name='event_source', field_type='STRING', mode='NULLABLE', description='Type of the worker the event originated from.'),
+  bigquery.SchemaField(name='event_class', field_type='STRING', mode='NULLABLE', description='Higher order category of event.'),
+  bigquery.SchemaField(name='event_type', field_type='STRING', mode='NULLABLE', description='The event type.'),
+  bigquery.SchemaField(name='session_id', field_type='STRING', mode='NULLABLE', description='The session ID, which is unique per client/server worker session.'),
+  bigquery.SchemaField(name='version_id', field_type='STRING', mode='NULLABLE', description='The version of the game\'s build or online service.'),
```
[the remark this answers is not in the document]
LoekL: Changed

services/python/analytics-pipeline/src/dataflow/common/classes.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class GetGcsFileList(beam.DoFn):
+
+    """ A custom Beam ParDo to generate a list of files that are present in
+    Google Cloud Storage (GCS). It takes path prefixes as elements (~ arguments).
+    """
+
+    def process(self, element):
+        from apache_beam.io.gcp import gcsio
```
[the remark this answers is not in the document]
LoekL: Importing stuff inside classes was a remnant of when I was using other pipelines as well, where `pipeline_options.view_as(SetupOptions).save_main_session = True` (in `p1_gcs_to_bq_backfill.py` always resulted in pickling errors).

If this had remained like this, I do believe for each element we would have re-imported everything, which wasn't perfect but manageable given the no. of elements passed to it. Now however I've moved them out of the class methods & added the import statements on top of `dataflow/common/classes.py` alongside the others. Worked when I tested it :)

services/python/analytics-pipeline/src/dataflow/p1_gcs_to_bq_backfill.py
LoekL: They're there so you can easily debug your pipeline if it starts to malfunction. 

By uncommenting all lines for either DataflowRunner or DirectRunner you would at each step write the intermediate results somewhere, to see what's going on.

I understand that I don't specifically say this anywhere. What do you think? Still remove or add this context somewhere, perhaps above the first pair of commented lines?

services/python/analytics-pipeline/src/function/main.py
LoekL: Good catch, another relic, this time from still using Python 2 earlier on. With Python 3 this is no longer required in this case: https://stackoverflow.com/questions/14083111/should-i-use-encoding-declaration-in-python-3

I've removed it & tested everything still works, which it does.

.gitignore
LoekL: Removed as per https://github.com/spatialos/online-services/pull/43#discussion_r311500097

services/python/analytics-pipeline/src/dataflow/p1_gcs_to_bq_backfill.py
JoelOtter: I can see the use but it leaves a bit of development overhead trying to keep those up to date. I think remove. If people are going to start debugging their pipelines they'll be able to work with the relevant documentation.

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior

I now specifically catch the `ValueError`

services/python/analytics-pipeline/src/dataflow/p1_gcs_to_bq_backfill.py
LoekL: Changed to `sys.exit(1)` based on the following statement: _"Some systems have a convention for assigning specific meanings to specific exit codes, but these are generally underdeveloped; Unix programs generally use 2 for command line syntax errors and 1 for all other kind of errors."_

source: https://docs.python.org/3/library/sys.html

JoelOtter: Yeah anything non-zero is fine.

services/python/analytics-pipeline/src/dataflow/p1_gcs_to_bq_backfill.py
LoekL: I now have:
```
# The below comment lines can help you if at any point you modify this pipeline and it starts to malfunction.
# By uncommenting for instance all Cloud-Debug lines the pipeline will write intermediate results of each step in files
# to your Google Cloud Storage bucket, so you can check what is going on at each step of your pipeline.
# fileListGcs | 'dumpGCSFileList' >> beam.io.WriteToText('gs://{gcs_bucket}/data_type=dataflow/batch/output/{job_name}/0_fileListGcs'.format(gcs_bucket = args.gcs_bucket, job_name = job_name)) # Cloud-Debug [when using DataflowRunner]
# fileListGcs | 'dumpGCSFileList' >> beam.io.WriteToText('local_debug/{job_name}/0_fileListGcs'.format(job_name = job_name)) # Local-Debug [when using DirectRunner]
```

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: Yeah already returning there makes total sense!

Later I rather have it `return None` than `raise Exception`, because I'd like to still store the event in BigQuery if only the timestamp is broken (can always easily remove afterwards if I want). Changed accordingly :)

services/python/analytics-pipeline/src/dataflow/p1_gcs_to_bq_backfill.py
LoekL: Missed your comment when I wrote the above reply. Removed all lines!

services/python/analytics-pipeline/src/dataflow/common/classes.py
```diff
@@ -0,0 +1,68 @@
...
+                gcs_uri_list_read = gcs.open(filename=file_path, mode='r').read().decode('utf-8').split('\n')
+                # With each **file** written into GCS by beam.io.WriteToText(), a PDone is returned & WriteToPubSub() is triggered,
+                # so we first remove this file from GCS to avoid duplicative results:
+                gcs_uri_list_delete = gcs.delete(path=file_path)
+
+                # Example GCS URI:
+                # gs://your-project-name-analytics/data_type=json/analytics_environment=function/event_category=scale-test/event_ds=2019-06-26/event_time=8-16/f58179a375290599dde17f7c6d546d78/2019-06-26T14:28:32Z-107087
+
+                for gcs_uri in gcs_uri_list_read:
+                    uri_prefix_offset = 5
```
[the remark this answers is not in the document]
LoekL: Changed to:

```
formatted_uri_list = gcs_uri.split("gs://", 1).pop().split('/')
gcs_bucket, name = formatted_uri_list[0], '/'.join(formatted_uri_list[1:])
```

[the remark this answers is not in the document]
LoekL: I am using:

```
gspath = gs://your-project-name-analytics/data_type=json/analytics_environment=function/event_category=scale-test/event_ds=2019-06-26/event_time=8-16/f58179a375290599dde17f7c6d546d78/2019-06-26T14:28:32Z-107087

filepath = /your-project-name-analytics/data_type=json/analytics_environment=function/event_category=scale-test/event_ds=2019-06-26/event_time=8-16/f58179a375290599dde17f7c6d546d78/2019-06-26T14:28:32Z-107087

object_location = data_type=json/analytics_environment=function/event_category=scale-test/event_ds=2019-06-26/event_time=8-16/f58179a375290599dde17f7c6d546d78/2019-06-26T14:28:32Z-107087

bucket_name = your-project-name-analytics
```

services/python/analytics-pipeline/src/dataflow/common/classes.py
```diff
@@ -0,0 +1,68 @@
...
+                    gcs_bucket = gcs_uri[uri_prefix_offset:].split('/')[0]
+                    name = '/'.join(gcs_uri[uri_prefix_offset:].split('/')[1:])
+                    if name and gcs_bucket:
+                        data = '{"name":"%s","bucket":"%s"}' % (name, gcs_bucket)
+                        future = client_ps.publish(topic, data=data.encode('utf-8'))
+                        yield (topic, data)
+
+            except Exception:
```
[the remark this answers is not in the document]
LoekL: I predominantly want to avoid the entire process quitting because a single file (part of a list of files) could not be downloaded. Instead I want it to continue to try the next file.

Following https://cloud.google.com/dataflow/docs/guides/logging I changed it to:
```python
except Exception as e:
    logging.info('Could not parse {gspath_list}: {e}'.format(gspath_list=gspath_list, e=e))
    continue
```

So whenever it excepts it'll write a log with all the details.

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: Actually, it would then turn into:

```python
try:
    # Try to split the path by key (indexing to 1 will fail if key not present in path):
    value = path.split(key)[1].split('/')[0]
    if key == 'event_ds=' and not validate_date(value):
        return None
    return value
except Exception:
    return None
```

Not sure if that's much better?

JoelOtter: I think it's better.

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: Yeah it's possible to just call the python script with `--analytics-environment=` & nothing afterwards, which causes it to be `None`.

What then happens:

```python
print(parse_analytics_environment(None))
([None], None)
```

So it would look for files like:

`gs://bucket-name-analytics/data_type=json/analytics_environment=/event_category=scale-test/event_ds=2019-06-26/event_time=8-16/f58179a375290599dde17f7c6d546d78/2019-06-26T14:28:32Z-107087` -> note the `analytics_environment=`. 

I think this is correct - but I hadn't thought of it, though..

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: Changed.

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: Hmm.. That passing it a list of lists would indeed mess things up.

For now changed it so that it:
a) Flattens the list first.
b) Casts all elements to strings (SQL does not handle a mix of types within an `IN ()` statement.

```python
def flatten_list(original_list):
    if isinstance(original_list, list):
        flattened_list = []
        for element in original_list:
            if isinstance(element,list): flattened_list.extend(flatten_list(element))
            else: flattened_list.append(element)
        return flattened_list
    else:
        raise TypeError('flatten_list() must be passed a list!')


def cast_elements_to_string(cast_list):
    if isinstance(cast_list, list):
        return [str(element) for element in cast_list]
    else:
        raise TypeError('cast_elements_to_string() must be passed a list!')


def convert_list_to_sql_tuple(filter_list):
    if isinstance(filter_list, list):
        return str(cast_elements_to_string(flatten_list(filter_list))).replace('[', '(').replace(']', ')')
    else:
        raise TypeError('convert_list_to_sql_tuple() must be passed a list!')

```

LoekL: I am probably going to take out the `cast_elements_to_string()` & `flatten_list()` calls from `convert_list_to_sql_tuple()`. It probably makes more sense we apply those on the argument we pass to `convert_list_to_sql_tuple()` instead.

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: Changed to:

```python
def get_dict_value(event_dict, *argv):

    """ This function takes as its first argument a dictionary, and afterwards any number of potenial
    keys to try to get a value for. The order of the potential keys matters, because as soon as any key
    yields a value it will return it (and quit). If none of the tried keys have an associated value, it will
    return None.
    """

    for arg in argv:
        value = event_dict.get(arg, None)
        if value:
            return value

    return None
```

services/python/analytics-pipeline/src/dataflow/common/classes.py
```diff
@@ -0,0 +1,68 @@
...
+    """ A custom Beam ParDo to notify a Pub/Sub Topic about the existence of files
+    in Google Cloud Storage (GCS).
+
+    It first reads GCS URI strings from a fileList stored in GCS, which should have happened in a
+    prior step. It then parses these GCS URIs by extracting the file name & bucket name, which it subsequently
+    uses as the payload for a Pub/Sub message to a particular Pub/Sub Topic.
+    """
+
+    def process(self, element, job_name, topic, gcp, gcs_bucket):
+        from apache_beam.io.gcp import gcsio
```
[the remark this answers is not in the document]
LoekL: Fixes as per https://github.com/spatialos/online-services/pull/43#discussion_r311516947

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: I am deliberately capturing the error message of the Exception, to subsequently write it alongside a bunch of meta-information into a debug table in BigQuery.

This way if I ever execute a backfill, it knows this file has failed before and it will omit it from the backfill. 

Makes sense?

services/python/analytics-pipeline/src/endpoint/main.py
```diff
@@ -0,0 +1,122 @@
...
+from random import randint
+
+client_storage = storage.Client.from_service_account_json(os.environ['SECRET_WRITER_JSON'])
+bucket = client_storage.get_bucket(os.environ['BUCKET_NAME'])
+
+try:
+    try:
+        subprocess.check_call('base64 --decode %s > /tmp/analytics-gcs-writer.p12' % os.environ['SECRET_WRITER_P12'], shell=True)
+    except Exception:
+        subprocess.call('cp %s /tmp/analytics-gcs-writer.p12' % os.environ['SECRET_WRITER_P12'], shell=True)
```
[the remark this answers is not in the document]
LoekL: I don't follow..

services/python/analytics-pipeline/src/endpoint/main.py
```diff
@@ -0,0 +1,122 @@
...
+        event_ds = request.args.get('ds', ds)
+        event_time = request.args.get('time', event_time)
+        session_id = request.args.get('session_id', 'session-id-not-available')
+
+        gcs_uri = 'data_type={data_type}/analytics_environment={analytics_environment}/event_category={event_category}/event_ds={event_ds}/event_time={event_time}/{session_id}/{ts_fmt}-{int}'
+
+        try:
+            payload = request.get_json(force=True)
+            gcs_destination = gcs_uri.format(data_type='json', analytics_environment=analytics_environment,
+              event_category=event_category, event_ds=event_ds, event_time=event_time, session_id=session_id, ts_fmt=ts_fmt, int=randint(100000, 999999))
```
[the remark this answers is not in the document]
LoekL: I've added a random integer as a suffix for the file name to avoid any potential collisions, if/when the endpoint ever tries to write two files with the exact same timestamp (which I use as its file name).

services/python/analytics-pipeline/src/endpoint/main.py
```diff
@@ -0,0 +1,122 @@
...
+
+            blob = bucket.blob(gcs_destination)
+            blob.upload_from_string('\n'.join(events), content_type='text/plain; charset=utf-8')
+
+            return jsonify({'code': 200, 'message': 'gs://{bucket_name}/{gcs_uri}'.format(bucket_name=bucket_name, gcs_uri=gcs_destination)})
+
+        except:
+            payload = request.get_data(as_text=True)
+            gcs_destination = gcs_uri.format(data_type='unknown', analytics_environment=analytics_environment, event_category=event_category,
+              event_ds=event_ds, event_time=event_time, session_id=session_id, ts_fmt=ts_fmt, int=randint(100000, 999999))
```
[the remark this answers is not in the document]
LoekL: I think I've done this in the mean-time:

```python
object_location_template = 'data_type={data_type}/analytics_environment={analytics_environment}/event_category={event_category}/event_ds={event_ds}/event_time={event_time}/{session_id}/{ts_fmt}-{int}'
object_location_json, object_location_json_raw, object_location_unknown = [object_location_template.format(data_type=data_type, analytics_environment=analytics_environment,
  event_category=event_category, event_ds=event_ds, event_time=event_time, session_id=session_id, ts_fmt=ts_fmt, int=randint(100000, 999999))
    for data_type in ['json', 'json_raw', 'unknown']]
```

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL:
```python
def cast_to_unix_timestamp(timestamp, timestamp_format_list):

    """ This function takes a timestamp and ensures a unix timestamp is returned,
    or None otherwise.

    An integer or float is returned as-is, whereas a timestamp in human readable
    string format is parsed using the provided timestamp format(s), verified to be valid
    & finally converted into a unix timestamp & returned.
    """

    # If timestamp is already in unix time, return as-is:
    if isinstance(timestamp, (int, float)):
        return timestamp

    # If timestamp is in human readable string format, try to parse using the given
    # formats, extract the unix timestamp if the timestamp is valid & return it:
    timestamp_list = []
    if isinstance(timestamp, str):
        for format in timestamp_format_list:
            try:
                return datetime.datetime.strptime(timestamp, format)
            except ValueError:
                continue
    return None
```

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
francissauch: can't this be done in the dockerfile ?

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
francissauch: I think something like that (from stack overflow)
`'.join(random.choices(string.ascii_uppercase + string.digits, k=6))`
would be a bit cleaner

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
francissauch: if someone gets access to the machine this things runs on and overrides the variable SECRET_WRITER_P12, they would be able to copy anything on the system to /tmp/analytics-gcs-writer.p12

That's an attack vector. Not sure how bad though.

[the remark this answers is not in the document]
LoekL: I might be confused, but in this case WRITER is not referring to be able to write to `/tmp/`, but that it's a key file of the Analytics GCS *Writer* service account. 

The environment variable name has since then been updated to better reflect this: `GOOGLE_SECRET_KEY_P12_ANALYTICS_GCS_WRITER`

If someone has access to the machine, and the key files, they'd basically be able to write whatever they want into a specific, single GCS bucket.

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
LoekL: Yeah we could probably get this done by configuring the `ENTRYPOINT` to be a bash script that first does the key conversion & afterwards boots the endpoint, but it comes at a small cost.

In the docs I plan to go step-by-step:

1. First triggering `endpoint/main.py` directly.
2. Containerizing `endpoint/main.py` & executing it as a standalone container locally.
3. Running the container together with Google's ESP container in a local pod using `docker-compose` (now a POST request for instance requires a valid API key).
4. Running the endpoint together with Google's ESP container in a cloud pod on GKE.

Obviously, in the first step there is no notion of a Dockerfile yet, so in order to get that working I'd have to tell people to first convert the `.p12` key file into `DER` using a few commands & afterwards pass it to the script via an additional environment variable . This versus the current approach where the script just handles everything for them, and requires just a single `GOOGLE_SECRET_KEY_...` environment variable. 

Alternatively, I just drop step 1) & move the conversion from `.p12` to `DER` to the Dockerfile's `ENTRYPOINT`.

Thoughts?

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
LoekL: Changed

services/python/analytics-pipeline/src/dataflow/common/classes.py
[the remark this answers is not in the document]
JoelOtter: Maybe log that as a warn rather than an info.

services/python/analytics-pipeline/src/dataflow/common/functions.py
JoelOtter: If you're ok with this that's fine, just wanted to make sure we covered all bases :)

services/python/analytics-pipeline/src/dataflow/common/functions.py
JoelOtter: I think so.

.gitignore
JoelOtter: This sounds like something we could just stick on internal wiki.

services/python/analytics-pipeline/src/dataflow/common/classes.py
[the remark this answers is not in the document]
LoekL: Changed to `logging.warning()`

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
LoekL: FIXED - All systems calls have been removed from `endpoint/main.py`.

Dockerfile `ENTRYPOINT` now triggers [entrypoint.sh](https://github.com/spatialos/online-services/blob/analytics-python/services/docker/analytics-endpoint/entrypoint.sh
) where these steps take place prior to booting up the endpoint.

services/python/analytics-pipeline/src/endpoint/main.py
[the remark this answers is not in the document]
LoekL: Fixed as per https://github.com/spatialos/online-services/pull/43#discussion_r312471168

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: I've made sure None is handled properly throughout the Dataflow script.

services/python/analytics-pipeline/src/dataflow/common/functions.py
LoekL: I've added and am now calling:

```python
def safe_convert_list_to_sql_tuple(filter_list):

    """ Safe version of convert_list_to_sql_tuple(), by first flattening the input list
    and then casting all its elements to strings, before proceeding with the conversion.
    """

    return convert_list_to_sql_tuple(cast_elements_to_string(flatten_list(filter_list)))
```

services/python/analytics-pipeline/src/function/main.py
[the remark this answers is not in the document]
LoekL: Updated to:

```python
def cast_object_to_string(object, object_type):

    """ Whenever object_type is either a list or dictionary, apply json.dumps()
    to cast it to a string, otherwise, apply str().
    """

    if object_type in [list, dict]:
        return json.dumps(object)
    else:
        return str(object)
```

I.e. you now pass the type alongside of it

PR #44, opened 2019-07-12
Analytics Docs

LoekL:
This PR contains all the docs + changes to existing docs revolving around the Analytics Pipeline.

PR #57, opened 2019-07-29
Base for analytics instrumentation system

ashleydavies:
This PR adds the foundations for instrumenting this repo for the reporting of analytics data.

The current state contains some basic classes for achieving this, with the main deficiencies being:
- Currently lacks a method of injecting the GCP key.
- Event categories are always empty.
- Messages are not batched, but instead all sent individually.

I intend to fix these in the next PR to avoid bloating this one too much.

### DIFF
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -74,14 +74,14 @@ namespace Common.Test
         private bool ExpectedMessage(HttpRequestMessage request)
         {
             var development = AnalyticsEnvironment.Development.ToString().ToLower();
-            
+
             Assert.IsInstanceOf<StringContent>(request.Content);
             if (request.Content is StringContent messageContent)
             {
                 dynamic content = JsonConvert.DeserializeObject(messageContent.ReadAsStringAsync().Result);
 
                 // TODO: Test versioning when it is added
-                Assert.AreEqual(content.eventEnvironment.Value, 
+                Assert.AreEqual(content.eventEnvironment.Value,
                     development);
                 Assert.AreEqual(content.eventIndex.Value, "0");
                 Assert.AreEqual(content.eventSource.Value, SourceVal);
@@ -113,7 +113,7 @@ namespace Common.Test
         public void SendAnalyticEventsToHttpsEndpoint()
         {
             HttpClient client = new HttpClient(_messageHandlerMock.Object);
-            AnalyticsSender.Build(new[] {$"--{EndpointName}", "https://example.com/"},
+            AnalyticsSender.Build(new[] { $"--{EndpointName}", "https://example.com/" },
                     AnalyticsEnvironment.Development, KeyVal, SourceVal, client)
                 .Send(ClassVal, TypeVal, new Dictionary<string, string>
                 {
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -66,19 +66,27 @@ namespace Common.Test
                 AnalyticsEnvironment.Development, ""));
         }
 
+        private const string SourceVal = "event_source_value";
+        private const string ClassVal = "event_class_value";
+        private const string TypeVal = "event_type_value";
+        private const string KeyVal = "fakeKey";
+
         private bool ExpectedMessage(HttpRequestMessage request)
         {
+            var development = AnalyticsEnvironment.Development.ToString().ToLower();
+            
             Assert.IsInstanceOf<StringContent>(request.Content);
             if (request.Content is StringContent messageContent)
             {
                 dynamic content = JsonConvert.DeserializeObject(messageContent.ReadAsStringAsync().Result);
 
                 // TODO: Test versioning when it is added
-                Assert.AreEqual(content.eventEnvironment.Value, "development");
+                Assert.AreEqual(content.eventEnvironment.Value, 
+                    development);
                 Assert.AreEqual(content.eventIndex.Value, "0");
-                Assert.AreEqual(content.eventSource.Value, "source");
-                Assert.AreEqual(content.eventClass.Value, "test");
-                Assert.AreEqual(content.eventType.Value, "send");
+                Assert.AreEqual(content.eventSource.Value, SourceVal);
+                Assert.AreEqual(content.eventClass.Value, ClassVal);
+                Assert.AreEqual(content.eventType.Value, TypeVal);
                 Assert.True(Guid.TryParse(content.sessionId.Value, out Guid _));
 
                 // Check the timestamp is within 5 seconds of now (i.e. roughly correct)
@@ -92,8 +100,8 @@ namespace Common.Test
             }
 
             var queryCollection = request.RequestUri.ParseQueryString();
-            Assert.AreEqual(queryCollection["key"], "fakeKey");
-            Assert.AreEqual(queryCollection["analytics_environment"], "development");
+            Assert.AreEqual(queryCollection["key"], KeyVal);
+            Assert.AreEqual(queryCollection["analytics_environment"], development);
             // TODO: Update with real category
             Assert.AreEqual(queryCollection["event_category"], "");
             Assert.True(Guid.TryParse(queryCollection["session_id"], out Guid _));
@@ -105,9 +113,9 @@ namespace Common.Test
         public void SendAnalyticEventsToHttpsEndpoint()
         {
             HttpClient client = new HttpClient(_messageHandlerMock.Object);
-            AnalyticsSender.Build(new[] { $"--{EndpointName}", "https://example.com/" },
-                    AnalyticsEnvironment.Development, "fakeKey", "source", client)
-                .Send("test", "send", new Dictionary<string, string>
+            AnalyticsSender.Build(new[] {$"--{EndpointName}", "https://example.com/"},
+                    AnalyticsEnvironment.Development, KeyVal, SourceVal, client)
+                .Send(ClassVal, TypeVal, new Dictionary<string, string>
                 {
                     {"dogs", "excellent"}
                 });
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -46,7 +46,7 @@ namespace Common.Test
         public void BuildRealAnalyticsSenderIfProvidedWithEndpoint()
         {
             Assert.IsInstanceOf<AnalyticsSender>(AnalyticsSender.Build(
-                new[] {$"--{EndpointName}", "https://example.com/"},
+                new[] { $"--{EndpointName}", "https://example.com/" },
                 AnalyticsEnvironment.Development, ""));
         }
 
@@ -54,7 +54,7 @@ namespace Common.Test
         public void FailToBuildIfHttpIsNotUsedWithoutInsecureEnabled()
         {
             Assert.Throws(typeof(ArgumentException),
-                () => AnalyticsSender.Build(new[] {$"--{EndpointName}", "http://example.com/"},
+                () => AnalyticsSender.Build(new[] { $"--{EndpointName}", "http://example.com/" },
                     AnalyticsEnvironment.Development, ""));
         }
 
@@ -62,14 +62,14 @@ namespace Common.Test
         public void AllowsHttpIfInsecureEndpointsEnabled()
         {
             Assert.IsInstanceOf<AnalyticsSender>(AnalyticsSender.Build(
-                new[] {$"--{EndpointName}", "http://example.com/", $"--{AllowInsecureEndpointName}"},
+                new[] { $"--{EndpointName}", "http://example.com/", $"--{AllowInsecureEndpointName}" },
                 AnalyticsEnvironment.Development, ""));
         }
 
         private bool ExpectedMessage(HttpRequestMessage request)
         {
             Assert.IsInstanceOf<FormUrlEncodedContent>(request.Content);
-            
+
             if (request.Content is FormUrlEncodedContent messageContent)
             {
                 NameValueCollection content = messageContent.ReadAsFormDataAsync().Result;
@@ -81,7 +81,7 @@ namespace Common.Test
                 Assert.AreEqual(content["eventClass"], "test");
                 Assert.AreEqual(content["eventType"], "send");
                 Assert.True(Guid.TryParse(content["sessionId"], out Guid _));
-                
+
                 // Check the timestamp is within 5 seconds of now (i.e. roughly correct)
                 long unixTimestampDelta = DateTimeOffset.UtcNow.ToUnixTimeSeconds()
                                           - long.Parse(content["eventTimestamp"]);
@@ -98,7 +98,7 @@ namespace Common.Test
             // TODO: Update with real category
             Assert.AreEqual(queryCollection["event_category"], "");
             Assert.True(Guid.TryParse(queryCollection["session_id"], out Guid _));
-            
+
             return request.Method == HttpMethod.Post;
         }
 
@@ -106,7 +106,7 @@ namespace Common.Test
         public void SendAnalyticEventsToHttpsEndpoint()
         {
             HttpClient client = new HttpClient(_messageHandlerMock.Object);
-            AnalyticsSender.Build(new[] {$"--{EndpointName}", "https://example.com/"},
+            AnalyticsSender.Build(new[] { $"--{EndpointName}", "https://example.com/" },
                     AnalyticsEnvironment.Development, "fakeKey", "source", client)
                 .Send("test", "send", new Dictionary<string, string>
                 {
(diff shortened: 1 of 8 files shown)


services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -0,0 +1,123 @@
...
+                });
+
+            // TODO: Verify contents of the message
+            _messageHandlerMock.Protected().Verify("SendAsync", Times.Exactly(1),
+                ItExpr.Is<HttpRequestMessage>(req => ExpectedMessage(req)),
+                ItExpr.IsAny<CancellationToken>());
+
+        }
+    }
+}
```
jared-improbable: Nit: add newlines at end of .cs files

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -0,0 +1,123 @@
...
+        private bool ExpectedMessage(HttpRequestMessage request)
+        {
+            Assert.IsInstanceOf<FormUrlEncodedContent>(request.Content);
+
+            if (request.Content is FormUrlEncodedContent messageContent)
+            {
+                NameValueCollection content = messageContent.ReadAsFormDataAsync().Result;
+
+                // TODO: Test versioning when it is added
+                Assert.AreEqual(content["eventEnvironment"], "development");
```
jared-improbable: Are there any constants that can be used for the keys and values in the content?

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        [Test]
+        public void SendAnalyticEventsToHttpsEndpoint()
+        {
+            HttpClient client = new HttpClient(_messageHandlerMock.Object);
+            AnalyticsSender.Build(new[] { $"--{EndpointName}", "https://example.com/" },
+                    AnalyticsEnvironment.Development, "fakeKey", "source", client)
+                .Send("test", "send", new Dictionary<string, string>
+                {
+                    {"dogs", "excellent"}
```
jared-improbable: Truth.

services/csharp/Common/Analytics/AnalyticsCommandLineArgs.cs
```diff
@@ -0,0 +1,17 @@
+using CommandLine;
+
+namespace Improbable.OnlineServices.Common.Analytics
+{
+    public class AnalyticsCommandLineArgs
+    {
+        public const string EndpointName = "analytics.endpoint";
+        public const string AllowInsecureEndpointName = "analytics.allowinsecureendpoint";
```
jared-improbable: Maybe add some dashes, as it's a bit of a mouthful.

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Threading.Tasks;
+using System.Web;
+using CommandLine;
+using Newtonsoft.Json;
+
+namespace Improbable.OnlineServices.Common.Analytics
+{
+    public enum AnalyticsEnvironment
+    {
+        Testing,
```
jared-improbable: Are these documented anywhere?

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        private AnalyticsSender(AnalyticsCommandLineArgs args, AnalyticsEnvironment environment,
+            string gcpKey, string eventSource, HttpClient httpClient)
+        {
+            _environment = environment;
+            _gcpKey = gcpKey;
+            _eventSource = eventSource;
+            _httpClient = httpClient;
+
+            Console.WriteLine($"Dispatching analytics to {args.Endpoint}");
```
jared-improbable: Library code probably shouldn't be jamming stuff into the console

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <param name="eventClass">A high level identifier for the event, e.g. deployment or gateway</param>
+        /// <param name="eventType">A more specific identifier for the event, e.g. `join`</param>
+        /// <param name="eventAttributes">A dictionary of k/v data about the event, e.g. user ID or queue duration</param>
+        public async Task Send(string eventClass, string eventType, Dictionary<string, string> eventAttributes)
+        {
+            BigInteger eventId;
+
+            lock (this)
+            {
+                eventId = _eventId++;
```
jared-improbable: Would a uint64 be ok for this, rather than BigInteger?
Also, `lock` is pretty heavy-weight, if you use uint64 then you can use more efficient constructs like https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=netcore-2.2#System_Threading_Interlocked_Increment_System_Int64__

Finally, if you must use `BigInteger`, then avoid `lock(this)` and instead make a new `eventIdLock` object to lock on instead.  (https://stackoverflow.com/questions/251391/why-is-lockthis-bad)

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -0,0 +1,127 @@
...
+                })
+            };
+
+            // TODO: Process response to handle failure / verify success
+            await _httpClient.PostAsync(builder.ToString(), new FormUrlEncodedContent(postParams));
+        }
+
+        private static string DictionaryToQueryString(Dictionary<string, string> urlParams)
+        {
+            List<string> entries = new List<string>(urlParams.Count);
```
jared-improbable: You could also LINQ-ify this all down to `return string.Join("&", entries.Select(kv => $"{HttpUtility.UrlEncode(kv.Key)}={HttpUtility.UrlEncode(kv.Value)}));`

services/csharp/Common/Analytics/IAnalyticsSender.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Improbable.OnlineServices.Common.Analytics
+{
+    public interface IAnalyticsSender
```
jared-improbable: Documentation strings?

services/csharp/Common/Analytics/NullAnalyticsSender.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// A black hole for analytics messages which does nothing; used as a default implementation for when analytics
+    /// is disabled.
+    /// </summary>
+    public class NullAnalyticsSender : IAnalyticsSender
+    {
+        public Task Send(string eventClass, string eventType, Dictionary<string, string> eventAttributes)
+        {
+            return null;
```
jared-improbable:
```suggestion
            return Task.CompletedTask;
```

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: They should be after https://github.com/spatialos/online-services/pull/44 is merged, and I'll add some documentation for this functionality once I've tightened the API up when I start integrating it with the existing services

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: I hadn't clocked precisely how big 64 bit integers are (quintillions, apparently) - yeah, that'll work. Thanks!

services/csharp/Common/Common.csproj
```diff
@@ -14,10 +14,12 @@
     <Copyright>2019 Improbable Worlds Ltd.</Copyright>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.5.0" />
     <PackageReference Include="Grpc.Core" Version="1.19.0" />
     <PackageReference Include="Improbable.SpatialOS.Platform" Version="14.0.0" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="../MemoryStore/MemoryStore.csproj" />
+    <ProjectReference Include="..\Base.Server\Base.Server.csproj" />
```
JoelOtter: I'm assuming this is here erroneously

services/csharp/Common.Test/AnalyticsShould.cs
ashleydavies: Have done with the value, but w.r.t the keys, I could make them a constant but to share them between the test and main codebase I think I'd need to make them public (or hack around it), and I anticipate they'll only ever be used just this once in the code - would you recommend doing so anyway?

jared-improbable: You could also make them `internal` and use the `InternalsVisibleToAttribute` (https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=netcore-2.2) to explicitly expose them to the `.Test` assembly, but I wouldn't really worry about it too much. If you change the strings, the tests will break, and then it's obvious and easily fixed. This was more of a "target of opportunity" question.

ashleydavies: Yeah these are tightly bound to the pipeline so unless they could be made global between both projects I think the redundancy in the testing may even be somewhat helpful to avoid errors. That said, that attribute sounds pretty useful & I'll keep it in mind for future - cheers!

services/csharp/Common/Common.csproj
ashleydavies: Whoops - yep

JoelOtter:
Should this PR target the `service-analytics` branch rather than `master`?

ashleydavies:
Can do if preferred, but these changes are independent of those & it would inflate https://github.com/spatialos/online-services/pull/43 quite a lot (I've got another PR incoming built on-top of this)

JoelOtter:
Yes, I think change the target. We're not shipping the analytics stuff at launch and I worry adding this to master could be a source of confusion for anyone looking through the code. `service-analytics` should be kept up-to-date with master anyway.

ashleydavies:
Ahh I just saw #43 is merging into `service-analytics` not master - makes sense, cheers!

services/csharp/Common.Test/Common.Test.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <PropertyGroup>
+        <TargetFramework>netcoreapp2.2</TargetFramework>
+
+        <IsPackable>false</IsPackable>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Moq" Version="4.10.1" />
+        <PackageReference Include="nunit" Version="3.11.0" />
+        <PackageReference Include="NUnit3TestAdapter" Version="3.11.0" />
```
JoelOtter: We're mostly using v`3.10.0` for this and `15.7.2` for the one below. Probably ought to be consistent (deployment pool is inconsistent, should fix in future)

services/csharp/Common/Common.csproj
```diff
@@ -14,6 +14,7 @@
     <Copyright>2019 Improbable Worlds Ltd.</Copyright>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.5.0" />
```
JoelOtter: `2.3.0`, unless there's some feature you need

services/csharp/Common/Analytics/IAnalyticsSender.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Improbable.OnlineServices.Common.Analytics
+{
+    /// <summary>
+    /// An interface for analytics senders to correspond to, used to facilitate the NullAnalyticsSender acting
+    ///   as a black hole for analytics.
```
JoelOtter: Nit: weird indentation here

services/csharp/Common/Analytics/IAnalyticsSender.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Improbable.OnlineServices.Common.Analytics
+{
+    /// <summary>
+    /// An interface for analytics senders to correspond to, used to facilitate the NullAnalyticsSender acting
```
JoelOtter: `correspond to` -> `implement` ?

services/csharp/Common.Test/Common.Test.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netcoreapp2.2</TargetFramework>
```
JoelOtter: `2.1`

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -0,0 +1,114 @@
...
+        Testing,
+        Development,
+        Staging,
+        Production,
+        Live,
+    }
+
+    public class AnalyticsSender : IAnalyticsSender
+    {
+        public static IAnalyticsSender Build(IEnumerable<string> args, AnalyticsEnvironment environment,
```
JoelOtter: It seems weird having this at the top of the class, even when it's static. Is that the standard?

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Improbable.OnlineServices.Common.Analytics;
+using Moq;
+using Moq.Protected;
+using Newtonsoft.Json;
+using NUnit.Framework;
+using static Improbable.OnlineServices.Common.Analytics.AnalyticsCommandLineArgs;
```
JoelOtter: Don't use this static - refer to the consts explicitly.

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -0,0 +1,130 @@
...
+
+namespace Common.Test
```
JoelOtter: Use the fully qualified namespace `Improbable.OnlineServices.Common.Test`

services/csharp/Common.Test/Common.Test.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
JoelOtter: You need to add a `RootNamespace` tag to this I think.

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -0,0 +1,114 @@
...
+            _environment = environment;
+            _gcpKey = gcpKey;
+            _eventSource = eventSource;
+            _httpClient = httpClient;
+            _endpoint = new Uri(args.Endpoint);
+
+            if (_endpoint.Scheme != Uri.UriSchemeHttps && !args.AllowInsecureEndpoints)
+            {
+                throw new ArgumentException(
+                    $"The endpoint provided uses {_endpoint.Scheme}, but only {Uri.UriSchemeHttps} is allowed. "
```
JoelOtter: What do these interpolated values look like? Are we verifying them anywhere?

ashleydavies: They're just the string literal for the protocol - http, https, etc. I'll verify it in the unit test

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: Not sure - I've always put static builder-y methods at the top, but I've got no objection to moving it. I assume you mean beneath the members?

JoelOtter: Yes. I'm assuming you can't do it in the constructor due to the static command line arguments stuff?

ashleydavies: Yeah - I've locked the constructor down because I want it to seamlessly return a NullAnalyticsSender when they don't provide an endpoint to avoid a bunch of redundant logic in every service.

Ideally I would put this Build method in the interface but sadly they're not allowed static methods. I considered a separate class, but I'm not sure it's worth it.

JoelOtter: Would be interested in @jared-improbable's thoughts here.

jared-improbable: I generally shove static methods towards the bottom, and use the StyleCop guidelines: https://stackoverflow.com/questions/2156129/is-there-a-standard-way-to-organize-methods-within-a-class?lq=1

services/csharp/Common.Test/Common.Test.csproj
ashleydavies: I think it was necessary in this case because we're using the Improbable.OnlineServices namespace - my bad!

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: Is it okay where it is now? I've moved it below the fields and properties but kept it above the constructor since it's pretty related in terms of flow. I think(?) it follows the StyleCop guide now

jared-improbable: Until we get automated tooling that can consistently enforce ordering, I really wouldn't sweat this too much.

PR #61, opened 2019-08-01
Analytics config

ashleydavies:
This PR adds analytics configuration file loading which allows event categories to be defined through a YAML file, as well as offering the ability to selectively disable events. There are other things we may want to add in future (such as reporting frequency, batch queue size, etc.) but these should be trivial to add at a later point.

Config files look like such:

```yaml
"party":
  "join":
    category: "function"

"user":
  "leave_party":
    category: "cold"
  "enter_party":
    disabled: true
```

I have some rough documentation for how this file operates, but I'm still working on how everything should fit together so I'd prefer to include that with the first PR that actually adds any instrumentation since it will have been finalised by then.

I've also made some improvements to the testing file since to amend mixing up the actual / expected parameter order.

In addition, I've gotten rid of the static `Build` method in `AnalyticsSender` and created a dedicated `Builder` class since I had ended up with a horrible mess of different constructors, with arguments almost exclusively to facilitate the test suite injecting a HttpClient.

### DIFF
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -110,7 +110,7 @@ namespace Improbable.OnlineServices.Common.Test
             Assert.AreEqual(KeyVal, queryCollection["key"]);
             Assert.AreEqual(development, queryCollection["analytics_environment"]);
             Assert.AreEqual(AnalyticsSender.DefaultEventCategory, queryCollection["event_category"]);
-            Assert.True(Guid.TryParse(queryCollection["session_id"], out Guid _));
+            Assert.True(Guid.TryParse((string) queryCollection["session_id"], out Guid _));
 
             return request.Method == HttpMethod.Post;
         }
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -16,9 +16,11 @@ namespace Improbable.OnlineServices.Common.Test
     public class AnalyticsShould
     {
         private Mock<HttpMessageHandler> _messageHandlerMock;
-        private string _gcpKey = "ABCDEF";
-        private string _eventSource = "event_source";
-        private AnalyticsEnvironment _devEnv = AnalyticsEnvironment.Development;
+        private const string SourceVal = "event_source_value";
+        private const string ClassVal = "event_class_value";
+        private const string TypeVal = "event_type_value";
+        private const string KeyVal = "gcp_key_value";
+        private const AnalyticsEnvironment DevEnv = AnalyticsEnvironment.Development;
 
         [SetUp]
         public void Setup()
@@ -41,7 +43,7 @@ namespace Improbable.OnlineServices.Common.Test
         public void BuildNullByDefault()
         {
             Assert.IsInstanceOf<NullAnalyticsSender>(
-                new AnalyticsSenderBuilder(_devEnv, _gcpKey, _eventSource).Build()
+                new AnalyticsSenderBuilder(DevEnv, KeyVal, SourceVal).Build()
             );
         }
 
@@ -49,7 +51,7 @@ namespace Improbable.OnlineServices.Common.Test
         public void BuildRealAnalyticsSenderIfProvidedWithEndpoint()
         {
             Assert.IsInstanceOf<AnalyticsSender>(
-                new AnalyticsSenderBuilder(_devEnv, _gcpKey, _eventSource)
+                new AnalyticsSenderBuilder(DevEnv, KeyVal, SourceVal)
                     .WithCommandLineArgs($"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/")
                     .Build()
             );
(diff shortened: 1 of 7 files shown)


services/csharp/Common/Analytics/Config/AnalyticsConfig.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            internal static readonly AnalyticsConfigEntry DefaultEntry = new AnalyticsConfigEntry();
+
+            public string Category { get; set; } = AnalyticsSender.DefaultEventCategory;
+            public bool Disabled { get; set; } = false;
+        }
+
+        public static async Task<AnalyticsConfig> FromFile(string filePath)
+        {
+            return new AnalyticsConfig(await File.ReadAllTextAsync(filePath));
```
jared-improbable: Why not just use `File.ReadAllText` since you're blocking on it anyway?
Then you can get rid of the `async` and the unneeded `Task` return.

services/csharp/Common/Analytics/Config/AnalyticsConfig.cs
```diff
@@ -0,0 +1,93 @@
...
+            var deserializer = new DeserializerBuilder()
+                .WithNamingConvention(new CamelCaseNamingConvention())
+                .Build();
+            try
+            {
+                _entries = deserializer.Deserialize<EntriesList>(contents) ?? new EntriesList();
+            }
+            catch (YamlException e)
+            {
+                throw new ArgumentException("Invalid YAML was provided", e);
```
jared-improbable: Does this really add any useful context? Consider removing this `try/catch`

services/csharp/Common.Test/Common.Test.csproj
```diff
@@ -9,10 +9,10 @@
...
     <ItemGroup>
         <PackageReference Include="Moq" Version="4.10.1" />
-        <PackageReference Include="nunit" Version="3.11.0" />
-        <PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
-        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.2" />
-        <PackageReference Include="System.Net.Http.Formatting.Extension" Version="5.2.3" />
+        <PackageReference Include="NUnit" Version="3.10.1" />
+        <PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.1.0" />
+        <PackageReference Include="Wdc.System.Net.Http.Formatting.NetStandard" Version="1.0.7" />
```
jared-improbable: What's this new package in service of?

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -16,6 +16,11 @@ namespace Improbable.OnlineServices.Common.Test
     public class AnalyticsShould
     {
         private Mock<HttpMessageHandler> _messageHandlerMock;
+        private const string SourceVal = "event_source_value";
+        private const string ClassVal = "event_class_value";
+        private const string TypeVal = "event_type_value";
+        private const string KeyVal = "gcp_key_value";
+        private const AnalyticsEnvironment DevEnv = AnalyticsEnvironment.Development;
```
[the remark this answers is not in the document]
ashleydavies: Just to make the tests more readable and less noisy since these values are used repeatedly; I can inline it if that's preferred?

services/csharp/Common.Test/Common.Test.csproj
ashleydavies: It handles the query string parsing in the test verifying the sent query contains the information it should (lines 109 -- 113)

This is a .NET Standard version rather than the .NET Framework version I had before; it functioned correctly but sent up multiple warnings whenever you'd run `dotnet test`

services/csharp/Common.Test/AnalyticsShould.cs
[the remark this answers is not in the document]
jared-improbable: I'd inline it, myself, but it's a nit overall

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -2,12 +2,17 @@
...
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
 using CommandLine;
+using CSharpx;
+using Improbable.OnlineServices.Common.Analytics.Config;
 using Newtonsoft.Json;
 
+[assembly: InternalsVisibleTo("Common.Test")]
```
JoelOtter: I'd like some justification for this. This seems like poor testing practice.

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -2,12 +2,17 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
 using CommandLine;
+using CSharpx;
```
JoelOtter: This looks like a mistake.

ashleydavies: Oops, I had fixed this in https://github.com/spatialos/online-services/pull/62/files but I'll move that change into this PR

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: This was only used to expose the `DefaultEventCategory` string to check defaults are set up correctly in the test suite; I'll move an explicit copy of the value into the testing suite instead

PR #62, opened 2019-08-02
Add batching features to the AnalyticsSender

ashleydavies:
This PR adds batching to the AnalyticsSender so that messages are not immediately sent but instead batched up and sent periodically or when the batch size becomes too large.

One evil of this PR is a Task.Delay in one of the tests, but I believe it's justified as it's only 20ms (the Common.Test suite as a whole takes ~700-900ms), I *think* NUnit handles async well, and it saves any extra complexity of making a wrapper around the Task library that can be injected. I'm happy to change this to another solution if preferred, though.

### DIFF
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -41,8 +41,6 @@ namespace Improbable.OnlineServices.Common.Test
                 }).Verifiable();
         }
 
-        private AnalyticsConfig _emptyConfig = new AnalyticsConfig("");
-
         [Test]
         public void BuildNullByDefault()
         {
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -116,7 +116,7 @@ namespace Improbable.OnlineServices.Common.Test
             Assert.AreEqual(KeyVal, queryCollection["key"]);
             Assert.AreEqual(development, queryCollection["analytics_environment"]);
             Assert.AreEqual(DefaultEventCategory, queryCollection["event_category"]);
-            Assert.True(Guid.TryParse((string) queryCollection["session_id"], out Guid _));
+            Assert.True(Guid.TryParse(queryCollection["session_id"], out Guid _));
 
             return request.Method == HttpMethod.Post;
         }
@@ -127,12 +127,13 @@ namespace Improbable.OnlineServices.Common.Test
             HttpClient client = new HttpClient(_messageHandlerMock.Object);
             new AnalyticsSenderBuilder(AnalyticsEnvironment.Development, KeyVal, SourceVal)
                 .WithCommandLineArgs( $"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/" )
+                .WithMaxQueueSize(1)
                 .With(client)
                 .Build()
                 .Send(ClassVal, TypeVal, new Dictionary<string, string>
                 {
                     { "dogs", "excellent" }
-                }, true);
+                });
 
             _messageHandlerMock.Protected().Verify("SendAsync", Times.Exactly(1),
                 ItExpr.Is<HttpRequestMessage>(req => SendAnalyticEventsToHttpsEndpointExpectedMessage(req)),
@@ -143,7 +144,7 @@ namespace Improbable.OnlineServices.Common.Test
         public async Task DispatchAnalyticsEventsForSameUriTogether()
         {
             HttpClient client = new HttpClient(_messageHandlerMock.Object);
-            IAnalyticsSender sender = new AnalyticsSenderBuilder(DevEnv, KeyVal, SourceVal)
+            IAnalyticsSender sender = new AnalyticsSenderBuilder(AnalyticsEnvironment.Development, KeyVal, SourceVal)
                 .WithMaxQueueSize(3)
                 .WithCommandLineArgs($"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/")
                 .With(client)
@@ -162,8 +163,8 @@ namespace Improbable.OnlineServices.Common.Test
         public async Task DispatchAnalyticsEventsAfterSomeTime()
         {
             HttpClient client = new HttpClient(_messageHandlerMock.Object);
-            IAnalyticsSender sender = new AnalyticsSenderBuilder(DevEnv, KeyVal, SourceVal)
-                .WithMaxQueueSize(3)
+            IAnalyticsSender sender = new AnalyticsSenderBuilder(AnalyticsEnvironment.Development, KeyVal, SourceVal)
+                .WithMaxQueueSize(5)
                 .WithMaxQueueTime(TimeSpan.FromMilliseconds(5))
                 .WithCommandLineArgs($"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/")
                 .With(client)
@@ -182,7 +183,7 @@ namespace Improbable.OnlineServices.Common.Test
         public async Task NotDispatchAnalyticsEventsWithoutTimeOrQueueSize()
         {
             HttpClient client = new HttpClient(_messageHandlerMock.Object);
-            IAnalyticsSender sender = new AnalyticsSenderBuilder(DevEnv, KeyVal, SourceVal)
+            IAnalyticsSender sender = new AnalyticsSenderBuilder(AnalyticsEnvironment.Development, KeyVal, SourceVal)
                 .WithMaxQueueSize(3)
                 .WithMaxQueueTime(TimeSpan.FromSeconds(5))
                 .WithCommandLineArgs($"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/")
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -54,7 +54,7 @@ namespace Improbable.OnlineServices.Common.Test
         {
             Assert.IsInstanceOf<AnalyticsSender>(
                 new AnalyticsSenderBuilder(AnalyticsEnvironment.Development, KeyVal, SourceVal)
-                    .WithCommandLineArgs( $"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/" )
+                    .WithCommandLineArgs($"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/")
                     .Build()
             );
         }
@@ -65,7 +65,7 @@ namespace Improbable.OnlineServices.Common.Test
             ArgumentException ex = Assert.Throws<ArgumentException>(
                 () =>
                     new AnalyticsSenderBuilder(AnalyticsEnvironment.Development, KeyVal, SourceVal)
-                        .WithCommandLineArgs( $"--{AnalyticsCommandLineArgs.EndpointName}", "http://example.com/" )
+                        .WithCommandLineArgs($"--{AnalyticsCommandLineArgs.EndpointName}", "http://example.com/")
                         .Build()
             );
 
@@ -126,7 +126,7 @@ namespace Improbable.OnlineServices.Common.Test
         {
             HttpClient client = new HttpClient(_messageHandlerMock.Object);
             new AnalyticsSenderBuilder(AnalyticsEnvironment.Development, KeyVal, SourceVal)
-                .WithCommandLineArgs( $"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/" )
+                .WithCommandLineArgs($"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/")
                 .WithMaxQueueSize(1)
                 .With(client)
                 .Build()
@@ -202,7 +202,7 @@ namespace Improbable.OnlineServices.Common.Test
         public void FallBackToDefaultConfigurationGracefully()
         {
             AnalyticsConfig config = new AnalyticsConfig("");
-            Assert.AreEqual(config.GetCategory("c", "t"), 
+            Assert.AreEqual(config.GetCategory("c", "t"),
                 DefaultEventCategory);
             Assert.True(config.IsEnabled("c", "t"));
         }
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -27,7 +27,7 @@ namespace Improbable.OnlineServices.Common.Test
         [SetUp]
         public void Setup()
         {
-            _messageHandlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
+            _messageHandlerMock = new Mock<HttpMessageHandler>(MockBehavior.Loose);
             _messageHandlerMock
                 .Protected()
                 .Setup<Task<HttpResponseMessage>>(
--- a/services/csharp/Common.Test/AnalyticsShould.cs
+++ b/services/csharp/Common.Test/AnalyticsShould.cs
@@ -40,6 +40,8 @@ namespace Improbable.OnlineServices.Common.Test
                     Content = new StringContent("")
                 }).Verifiable();
         }
+        
+        private AnalyticsConfig _emptyConfig = new AnalyticsConfig("");
 
         [Test]
         public void BuildNullByDefault()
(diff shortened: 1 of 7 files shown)


services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
```diff
@@ -9,6 +9,17 @@ namespace Improbable.OnlineServices.Common.Analytics
...
+        /// <summary>
+        /// Maximum size of the event queue before all events within it are dispatched
+        /// </summary>
+        private int _maxEventQueueSize = 10;
+
+        /// <summary>
+        /// Maximum time an event should wait in the queue before being dispatched to the endpoint.
+        /// May be longer if an event is added while previous events are being dispatched.
+        /// </summary>
+        private int _maxQueueTimeMs = 2500;
```
jared-improbable: A `TimeSpan` may be more appropriate here

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -31,32 +30,46 @@ public class AnalyticsSender : IAnalyticsSender
...
+                while (true)
+                {
+                    await DispatchEventQueue();
+                    await Task.Delay(maxEventQueueDeltaMs, _queueTimedDispatchCancellationToken);
+                }
+            }, _queueTimedDispatchCancellationToken, TaskCreationOptions.LongRunning, TaskScheduler.Default);
         }
 
-        public async Task Send(string eventClass, string eventType, Dictionary<string, string> eventAttributes)
+        public async Task Send(string eventClass, string eventType, Dictionary<string, string> eventAttributes,
```
jared-improbable: Adding a `bool` parameter is usually a code smell.
How about refactoring this into `Send` and `SendAsync`? If that doesn't work, then consider adding an `enum`, which greatly helps readability at the call site.

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -82,9 +95,63 @@ public async Task Send(string eventClass, string eventType, Dictionary<string, s
...
+
+            // We only need to lock dequeue operations so we can ensure batches are sent together even if
+            // both conditions occur at the same time - the time elapses and the queue fills at the same time.
+            // In the event of a second thread trying to dispatch at the same time as another is, we just fall through
+            // with the empty uriMap
+            if (Monitor.TryEnter(_queuedRequests))
+            {
+                try
+                {
+                    if (_queuedRequests.Count == 0) return;
```
jared-improbable:
```suggestion
                    if (_queuedRequests.Count == 0) 
                    {
                    	return;
                    }
```

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -82,9 +95,63 @@ public async Task Send(string eventClass, string eventType, Dictionary<string, s
...
+                finally
+                {
+                    Monitor.Exit(_queuedRequests);
+                }
+            }
+
+            var enumerable = uriMap.Select(
+                kvp => SendData(kvp.Key, new StringContent(string.Join("\n", kvp.Value)))
+            );
+            foreach (var result in enumerable) await result;
```
jared-improbable:
```suggestion
            Task.WaitAll(enumerable.ToArray());
```

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -82,9 +95,63 @@ public async Task Send(string eventClass, string eventType, Dictionary<string, s
...
-                new StringContent(JsonConvert.SerializeObject(postParams)));
+            var uri = builder.Uri;
+
+            if (immediateSend)
+            {
+                await SendData(uri, new StringContent(JsonConvert.SerializeObject(postParams)));
+            }
+            else
+            {
+                bool shouldDispatchQueue = false;
```
jared-improbable:
```suggestion
```

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -82,9 +95,63 @@ public async Task Send(string eventClass, string eventType, Dictionary<string, s
...
+
+                _queuedRequests.Enqueue((uri, JsonConvert.SerializeObject(postParams)));
+
+                if (_queuedRequests.Count >= _maxEventQueueSize)
+                {
+                    shouldDispatchQueue = true;
+                }
+
+                if (shouldDispatchQueue) await DispatchEventQueue();
```
jared-improbable:
```suggestion
```

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -82,9 +95,63 @@ public async Task Send(string eventClass, string eventType, Dictionary<string, s
...
+            }
+            else
+            {
+                bool shouldDispatchQueue = false;
+
+                _queuedRequests.Enqueue((uri, JsonConvert.SerializeObject(postParams)));
+
+                if (_queuedRequests.Count >= _maxEventQueueSize)
+                {
+                    shouldDispatchQueue = true;
```
jared-improbable:
```suggestion
                    await DispatchEventQueue();
```

services/csharp/Common/Analytics/NullAnalyticsSender.cs
```diff
@@ -9,7 +9,7 @@ namespace Improbable.OnlineServices.Common.Analytics
     /// </summary>
     public class NullAnalyticsSender : IAnalyticsSender
     {
-        public Task Send(string eventClass, string eventType, Dictionary<string, string> eventAttributes)
+        public Task Send(string eventClass, string eventType, Dictionary<string, string> eventAttributes, bool imm)
```
jared-improbable:
```suggestion
        public Task Send(string eventClass, string eventType, Dictionary<string, string> eventAttributes, bool immediate)
```

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -82,9 +95,63 @@ public async Task Send(string eventClass, string eventType, Dictionary<string, s
...
+
+                if (shouldDispatchQueue) await DispatchEventQueue();
+            }
+        }
+
+        private async Task DispatchEventQueue()
+        {
+            Dictionary<Uri, List<string>> uriMap = new Dictionary<Uri, List<string>>();
+
+            // We only need to lock dequeue operations so we can ensure batches are sent together even if
```
jared-improbable: Is there a requirement that already-queued data be sent before batched data?

If there's not, then I think that this should be greatly simplified, and a `SendBatch` API added instead. I'd definitely prefer a more explicit interface.

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: 😨 oops, I have no idea why I didn't spot this

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: Sorry, I'm a little confused - what do you mean by batched data? For me, already-queued and batched mean the same thing in this context.

ashleydavies:
@jared-improbable All comments addressed now I think; it's dramatically simplified:

1. Doesn't use monitors anymore
1. No `immediate` sending mode; all requests get queued unconditionally (couldn't think of any good reason why not)

Sorry for the delay; the rebase had dozens of conflicts for some reason

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -27,35 +27,65 @@ public class AnalyticsSender : IAnalyticsSender
...
 
+        private readonly ConcurrentQueue<(Uri uri, string content)> _queuedRequests =
+            new ConcurrentQueue<(Uri, string)>();
+
         private long _eventId;
 
-        internal AnalyticsSender(Uri endpoint, AnalyticsConfig config, AnalyticsEnvironment environment,
-            string gcpKey, string eventSource, HttpClient httpClient)
+        private string EnvironmentString => _environment.ToString().ToLower();
+        private bool QueueFull => _queuedRequests.Count >= _maxEventQueueSize;
```
jared-improbable:
```suggestion
        private bool IsQueueFull => _queuedRequests.Count >= _maxEventQueueSize;
```

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -131,7 +136,65 @@ public void SendAnalyticEventsToHttpsEndpoint()
...
+                .WithCommandLineArgs($"--{AnalyticsCommandLineArgs.EndpointName}", "https://example.com/")
+                .With(client)
+                .Build();
+
+            await sender.Send(ClassVal, TypeVal, new Dictionary<string, string>());
+            await sender.Send("class-val-2", "type-val-2", new Dictionary<string, string>());
+
+            _messageHandlerMock.Protected().Verify("SendAsync", Times.Exactly(0),
+                ItExpr.IsAny<HttpRequestMessage>(),
                 ItExpr.IsAny<CancellationToken>());
```
jared-improbable: How about tests for sending when the queue is full?
Any more tests to add?

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -27,35 +27,65 @@ public class AnalyticsSender : IAnalyticsSender
...
         private readonly HttpClient _httpClient;
 
+        private readonly ConcurrentQueue<(Uri uri, string content)> _queuedRequests =
+            new ConcurrentQueue<(Uri, string)>();
+
         private long _eventId;
 
-        internal AnalyticsSender(Uri endpoint, AnalyticsConfig config, AnalyticsEnvironment environment,
-            string gcpKey, string eventSource, HttpClient httpClient)
+        private string EnvironmentString => _environment.ToString().ToLower();
```
jared-improbable:
```suggestion
        private string CanonicalEnvironment => _environment.ToString().ToLower();
```

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -27,35 +27,65 @@ public class AnalyticsSender : IAnalyticsSender
         private readonly Uri _endpoint;
```
jared-improbable: I can't remember if we discussed making this implement `IDisposable` and cancelling the scraping task in `Dispose`

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -66,21 +96,42 @@ public async Task Send(string eventClass, string eventType, Dictionary<string, s
...
+
+            var enumerable = uriMap.Select(
+                kvp => SendData(kvp.Key, new StringContent(string.Join("\n", kvp.Value)))
+            );
+            Task.WaitAll(enumerable.ToArray<Task>());
+        }
+
+        private Task<HttpResponseMessage> SendData(Uri uri, StringContent content)
+        {
+            // TODO: Process response to handle failure / verify success; possibly falling back to alternative endpoint
```
jared-improbable: Is there a task tracked for this work?

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -27,35 +27,65 @@ public class AnalyticsSender : IAnalyticsSender
...
+            {
+                while (true)
+                {
+                    await DispatchEventQueue();
+                    await Task.Delay(maxEventQueueDelta, _queueTimedDispatchCancellationToken);
+                }
+            }, _queueTimedDispatchCancellationToken, TaskCreationOptions.LongRunning, TaskScheduler.Default);
         }
 
         public async Task Send(string eventClass, string eventType, Dictionary<string, string> eventAttributes)
```
jared-improbable: This doesn't really need to return a `Task`, there's nothing the user can do with it.

services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
```diff
@@ -58,6 +70,24 @@ public AnalyticsSenderBuilder With(AnalyticsConfig config)
             return this;
         }
 
+        /// <summary>
+        /// Sets the maximum size the analytics event queue should reach before the queue is dispatched
```
jared-improbable: Nit: full stop at end of comment (and again below)

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: Isn't it valuable for exception handling and waiting? Not fully clued in on the subtleties though, so happy to remove if not

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -27,35 +27,65 @@ public class AnalyticsSender : IAnalyticsSender
         private readonly Uri _endpoint;
         private readonly AnalyticsConfig _config;
         private readonly AnalyticsEnvironment _environment;
+        private readonly CancellationToken _queueTimedDispatchCancellationToken = new CancellationToken();
+        private readonly int _maxEventQueueSize;
```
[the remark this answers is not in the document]
ashleydavies: The initial idea I believe was to ensure that if any random spikes are built up then not too many analytics are sent at once, but discussed with Loek and it seems like we can live without this unless it becomes a need later on

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: We did but the intended use of this service is to exist for the lifecycle of the program. I'll implement it anyway to ensure the tests are cleaning up & it's only a couple of lines

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -104,41 +112,88 @@ private bool ExpectedMessage(HttpRequestMessage request)
...
             Assert.AreEqual(DefaultEventCategory, queryCollection["event_category"]);
-            Assert.True(Guid.TryParse((string) queryCollection["session_id"], out Guid _));
+            Assert.True(Guid.TryParse(queryCollection["session_id"], out var _));
 
             return request.Method == HttpMethod.Post;
         }
 
         [Test]
-        public void SendAnalyticEventsToHttpsEndpoint()
+        public async Task SendAnalyticEventsToHttpsEndpoint()
```
jared-improbable: `async void` is fine

ashleydavies: It seems like NUnit 3 doesn't support async void?

services/csharp/Common/Analytics/AnalyticsSender.cs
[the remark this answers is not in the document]
LoekL: @jared-improbable, the idea is to:
- Never end up with massive batches -> why we have the batch size constraint.
- Never end up with large latency -> why we have the time constraint.

I'm potentially planning to write events as GZIP'ed JSON files in GCS to align with observability, which does this to save storage space (which is more of an issue for them than for us). This will however result in the loss of being able to read in individual files in parallel with BigQuery (so we can still read many files at once, but can no longer speed up parsing a single file using parallelism), when using GCS as a federated data source for an external table.

If I end up with say 100k gzipped files that each contain 100 events, and 1 gzipped file with 10k events, this single 10k file will stall my entire query, which is what I want to avoid, hence the batch size restriction.

Hope this makes sense, do you think with this context it makes sense to keep the batch size restriction?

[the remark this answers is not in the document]
jared-improbable: Within that context the constraint makes sense and seems worth keeping; thanks for the extra detail!
I like to try to keep things as minimal as possible, so I tend to question things just to make sure :)

services/csharp/Common/Analytics/AnalyticsSender.cs
ashleydavies: Added OS-480

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -104,41 +112,88 @@ private bool ExpectedMessage(HttpRequestMessage request)
...
+                    .Build())
+            {
+                sender.Send(ClassVal, TypeVal, new Dictionary<string, string>());
+                sender.Send("class-val-2", "type-val-2", new Dictionary<string, string>());
+
+                _messageHandlerMock.Protected().Verify("SendAsync", Times.Exactly(0),
+                    ItExpr.IsAny<HttpRequestMessage>(),
+                    ItExpr.IsAny<CancellationToken>());
+            }
         }
```
jared-improbable: Just re-upping the need for testing queue limit sending tests.

ashleydavies: Whoops - I had one and forgot to re-add it with the others. Cheers!

PR #66, opened 2019-08-06
Add exception handling strategies to analytics sender

ashleydavies:
This PR tackles the issue that, ideally, sending analytics should be low-effort. Users shouldn't have to wrap every analytics.Send(...) call with a near-identical try-catch blocks to avoid their application crashing, since it disincentives analytics sending.

Added:

* System of `DispatchExceptionStrateg[ies]` which are pretty much a layer over a simple handler (we could also use an action and allow an arbitrary lambda - open to feedback here, though it seems sensible to me to at least provide some pre-defined options)
* A log handler, which I intend to use to instrument the existing services.
* A simple handler for just re-throwing the exception so the user deals with it - essentially a wrapper around the old behaviour. This is enabled by default.
* Corresponding unit tests to verify that exceptions are routed through the system correctly and that both handlers function as expected.

### DIFF
--- a/services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
+++ b/services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Net.Http;
 using CommandLine;
 using Improbable.OnlineServices.Common.Analytics.Config;
+using Improbable.OnlineServices.Common.Analytics.ExceptionHandlers;
 
 namespace Improbable.OnlineServices.Common.Analytics
 {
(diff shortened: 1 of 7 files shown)


services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
```diff
@@ -70,6 +72,16 @@ public AnalyticsSenderBuilder With(AnalyticsConfig config)
             return this;
         }
 
+        /// <summary>
+        /// Sets the strategy for handling HTTP exceptions during analytic dispatch. Silently ignores null strategies
+        /// in favour of the existing strategy.
+        /// </summary>
+        public AnalyticsSenderBuilder With(IDispatchExceptionStrategy strategy)
+        {
+            _dispatchExceptionStrategy = strategy ?? _dispatchExceptionStrategy;
```
jared-improbable: I would find this slightly surprising - how about throwing `ArgumentNullException` instead?

PR #70, opened 2019-08-07
Various changes to analytics in advance of instrumenting Party/

ashleydavies:
This PR makes a variety of changes which made instrumenting Party (next PR) more straightforward. My focus was on minimising duplication where possible to make the user-level instrumentation as minimal as possible.

Main changes:
* Accept generic type T to Send to allow non-string data e.g. lists to be encoded into JSON correctly. For most events, T = string.
* Delete `development` environment
* Add GcpKeyPath and environment command line flags
* Split `Send` into `Send` and `SendAsync`.
* Add a `AnalyticsSenderClassWrapper` which wraps around `IAnalyticsSender` and provides some utility methods to avoid having to pass in `eventClass` every time, since it's typically the same.
   * This could easily be part of the ordinary interface with a specific default provided, but tricky behaviour if users forget to provide a default, and I don't want to require one if possible
   * This also allows the same analytics sender to have different defaults in different places - if you have e.g. two 'servers' sharing a single process, they can share a single analytics sender wrapped by different class wrappers.

### DIFF
--- a/services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
+++ b/services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Net.Http;
 using CommandLine;
+using Google.Apis.Logging;
 using Improbable.OnlineServices.Common.Analytics.Config;
 using Improbable.OnlineServices.Common.Analytics.ExceptionHandlers;
 
@@ -21,11 +23,11 @@ namespace Improbable.OnlineServices.Common.Analytics
         /// </summary>
         private TimeSpan _maxQueueTime = TimeSpan.FromMilliseconds(2000);
 
-        private readonly AnalyticsEnvironment _environment;
-        private readonly string _gcpKey;
+        private AnalyticsEnvironment _environment;
         private readonly string _eventSource;
         private AnalyticsConfig _config;
         private bool _allowUnsafeEndpoints;
+        private string _gcpKey;
         private HttpClient _httpClient = new HttpClient();
         private Uri _endpoint;
         private IDispatchExceptionStrategy _dispatchExceptionStrategy = new RethrowExceptionStrategy();
@@ -41,11 +43,16 @@ namespace Improbable.OnlineServices.Common.Analytics
             _eventSource = eventSource;
         }
 
+        public AnalyticsSenderBuilder(string eventSource)
+        {
+            _eventSource = eventSource;
+        }
+
         public IAnalyticsSender Build()
         {
             _config = _config ?? new AnalyticsConfig();
 
-            if (_endpoint != null)
+            if (_endpoint != null && !string.IsNullOrEmpty(_gcpKey))
             {
                 if (_endpoint.Scheme != Uri.UriSchemeHttps && !_allowUnsafeEndpoints)
                 {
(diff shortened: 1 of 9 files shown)


services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
```diff
@@ -119,6 +127,12 @@ public AnalyticsSenderBuilder WithCommandLineArgs(IEnumerable<string> args)
                         _endpoint = new Uri(parsedArgs.Endpoint);
                     }
 
+                    if (!string.IsNullOrEmpty(parsedArgs.GcpKeyPath))
+                    {
+                        _gcpKey = File.ReadAllText(parsedArgs.GcpKeyPath);
```
jared-improbable: Depending on where that key file comes from, adding a `.Trim()` to the content read from disk may be a good idea.

services/csharp/Common/Analytics/AnalyticsSenderClassWrapper.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Threading.Tasks;
+
+namespace Improbable.OnlineServices.Common.Analytics
+{
+    public class AnalyticsSenderClassWrapper : IAnalyticsSender
+    {
+        private readonly IAnalyticsSender _wrapped;
+        private readonly string _class;
+
+        public AnalyticsSenderClassWrapper(IAnalyticsSender wrapped, string @class)
```
jared-improbable: `@class` -> `className`?

services/csharp/Common/Analytics/AnalyticsSenderClassWrapper.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public Task SendAsync<T>(string eventType, Dictionary<string, T> eventAttributes)
+        {
+            return SendAsync<T>(_class, eventType, eventAttributes);
+        }
+
+        public void Dispose()
+        {
+            _wrapped.Dispose();
```
jared-improbable: It feels a bit odd for this to take ownership, but I guess I can't see the call sites to get a feel for it, either.

services/csharp/Common/Analytics/IAnalyticsSender.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
```
jared-improbable: Is this needed?

services/csharp/Common/Analytics/AnalyticsSenderClassWrapper.cs
ashleydavies: Actually this is a good point - my intention was for multiple of these objects to be able to be used to wrap a single analytics sender, so this is definitely wrong as-is - thanks!

services/csharp/Common/Analytics/AnalyticsSenderClassWrapper.cs
```diff
@@ -0,0 +1,44 @@
...
+            Send(_eventClass, eventType, eventAttributes);
+        }
+
+        public Task SendAsync<T>(string eventType, Dictionary<string, T> eventAttributes)
+        {
+            return SendAsync(_eventClass, eventType, eventAttributes);
+        }
+
+        /// <summary>
+        /// The owner is the analytics sender is still responsible for disposing it.
```
jared-improbable:
```suggestion
        /// The owner of the analytics sender is still responsible for disposing it.
```

PR #71, opened 2019-08-07
More pre-instrumentation changes to client-side Analytics API

ashleydavies:
This PR adds analytics instrumentation to the Party project - to both the invite and party services.

Currently tests are just given NullAnalyticsSenders so they don't interfere, but I intend to add tests prior to merging; mostly after an OK on what's here now before spending too much effort on that.


services/csharp/Party.Test/CreateInviteShould.cs
```diff
@@ -38,7 +39,7 @@ public void SetUp()
 
             var memoryStoreClientManager = new Mock<IMemoryStoreClientManager<IMemoryStoreClient>>(MockBehavior.Strict);
             memoryStoreClientManager.Setup(manager => manager.GetClient()).Returns(_mockMemoryStoreClient.Object);
-            _inviteService = new InviteServiceImpl(memoryStoreClientManager.Object);
+            _inviteService = new InviteServiceImpl(memoryStoreClientManager.Object, new NullAnalyticsSender().WithEventClass(""));
```
jared-improbable: `.WithEventClass("")` seems redundant to need to include.

services/csharp/Party/InviteServiceImpl.cs
```diff
@@ -18,13 +19,17 @@ namespace Party
     public class InviteServiceImpl : InviteService.InviteServiceBase
     {
         private readonly IMemoryStoreClientManager<IMemoryStoreClient> _memoryStoreClientManager;
+        private readonly AnalyticsSenderClassWrapper _analytics;
 
-        public InviteServiceImpl(IMemoryStoreClientManager<IMemoryStoreClient> memoryStoreClientManager)
+        public InviteServiceImpl(IMemoryStoreClientManager<IMemoryStoreClient> memoryStoreClientManager,
+            AnalyticsSenderClassWrapper analytics)
```
jared-improbable: Wouldn't `IAnalyticsSender` make more sense?

services/csharp/Party/PartyServiceImpl.cs
```diff
@@ -137,23 +162,29 @@ public override async Task<JoinPartyResponse> JoinParty(JoinPartyRequest request
...
+                        "The player is not invited to this party"));
                 }
-                var invited = (await Task.WhenAll(playerInvites.InboundInviteIds
+
+                var invites = (await Task.WhenAll(playerInvites.InboundInviteIds
                         .Select(invite => memClient.GetAsync<Invite>(invite))))
-                        .Where(invite =>
+                    .Where(invite =>
+                    {
+                        if (invite == null)
```
jared-improbable: Feels weird to log in a `Where` statement

ashleydavies: I didn't add this, I think my editor just changed the indentation level as I modified the return type to re-use it

services/csharp/Common/Analytics/AnalyicsSenderBuilder.cs
```diff
@@ -135,7 +133,7 @@ public AnalyticsSenderBuilder WithCommandLineArgs(IEnumerable<string> args)
 
                     if (!string.IsNullOrEmpty(parsedArgs.Environment))
                     {
-                        bool result = Enum.TryParse(parsedArgs.Environment, out _environment);
+                        bool result = Enum.TryParse(parsedArgs.Environment, true, out _environment);
```
jared-improbable: Can you please use a named parameter for the `bool` argument?

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -149,7 +149,10 @@ private async Task DispatchEventQueue()
             try
             {
                 var enumerable = uriMap.Select(kvp =>
-                    _httpClient.PostAsync(kvp.Key, new StringContent(string.Join("\n", kvp.Value))));
+                    _httpClient.PostAsync(kvp.Key,
+                        new StringContent("[" + string.Join(",", kvp.Value) + "]")
```
jared-improbable:
```suggestion
                        new StringContent($"[{string.Join(",", kvp.Value)}]")
```

services/csharp/Party/Program.cs
```diff
@@ -33,7 +35,7 @@ public static void Main(string[] args)
             ThreadPool.GetMaxThreads(out var workerThreads, out var ioThreads);
             ThreadPool.SetMinThreads(workerThreads, ioThreads);
 
-            Parser.Default.ParseArguments<PartyServerCommandLineArgs>(args)
+            new Parser(w => w.IgnoreUnknownArguments = true).ParseArguments<PartyServerCommandLineArgs>(args)
```
jared-improbable: Why? This could lead to odd bugs, potentially.

ashleydavies: This allows the `AnalyticsCommandLineArgs` to be disjoint. I agree this is really gross, but I'm a bit stuck on how to improve it without a feature like mixins.

We could make a hierarchy like CommandLineArgs > AnalyticsCommandLineArgs > XServerCommandLineArgs, but that feels a little gross too - maybe less so?

ashleydavies: One alternative is moving all analytics configuration into a config file and having any server with analytics handle accepting an argument to a path to that file?

jared-improbable: The parser does work with `[Option]`-annotated `interface`s - I've used that as a way of "mixing-in" multiple sets of options.
For example: https://github.com/spatialos/csharp-worker-template/blob/master/Improbable/Postgres/Improbable.Postgres/IPostgresOptions.cs

ashleydavies: Sounds good, I'll do that!

ashleydavies:
I'm going to rip out the Party/ instrumentation from this PR since this PR has a lot of irrelevant stuff which is probably best kept alone for now

I'll send up the Party/ stuff in another PR shortly after I land this

ashleydavies:
@jared-improbable Migrated everything to use interfaces now, thanks for the tip! - changes of note:

* Verify methods moved to be extension methods in the relevant `Program` files
* Moved consts to a static class for the analytics argument class, and imported it static in the test suite to avoid major repetition
* I moved the instrumentation of Party out of this PR to put in a separate PR as I didn't want to conflate all these changes together and wanted to land this earlier as I believe @rebekahmullen is working on instrumenting Playfab auth

ashleydavies:
Ah, looks like it can't actually create an instance the interfaces (I assumed it would do some magic with reflection); I'll look further into what you did and see what needs to be changed to get this working.

PR #74, opened 2019-08-09
Instrument party and invite services

ashleydavies:
Nothing new in this PR from what was previously in the other PR except for how the command line arg class is set up.

There is a little redundancy in PartyServiceCommandLineArgs now, but some nice benefits:

* Don't need to parse twice
* Can still forbid unknown arguments

### DIFF
--- a/services/csharp/Party/PartyServerCommandLineArgs.cs
+++ b/services/csharp/Party/PartyServerCommandLineArgs.cs
@@ -1,10 +1,11 @@
 using System;
 using CommandLine;
 using Improbable.OnlineServices.Base.Server;
+using Improbable.OnlineServices.Common.Analytics;
 
 namespace Party
 {
-    public class PartyServerCommandLineArgs : CommandLineArgs
+    public class PartyServerCommandLineArgs : CommandLineArgs, IAnalyticsCommandLineArgs
     {
         [Option("redis_connection_string", HelpText = "The connection string for redis (host:port)",
             Default = "127.0.0.1:6379")]
@@ -30,5 +31,12 @@ namespace Party
                 throw new ArgumentException("DefaultMinMembers cannot be negative");
             }
         }
+
+        // Analytics args
+        public string Endpoint { get; set; }
+        public bool AllowInsecureEndpoints { get; set; }
+        public string ConfigPath { get; set; }
+        public string GcpKeyPath { get; set; }
+        public string Environment { get; set; }
     }
 }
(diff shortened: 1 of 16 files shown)


services/csharp/Party/PartyServerCommandLineArgs.cs
```diff
@@ -30,5 +31,12 @@ public void Validate()
                 throw new ArgumentException("DefaultMinMembers cannot be negative");
             }
         }
+
+        // Analytics args
```
jared-improbable: One thing I kind of like (though I'm totally inconsistent about doing it) is surrounding interface implementations with
```
#region IAnalyticsCommandLineArgs
...
#endregion IAnalyticsCommandLineArgs
```

services/csharp/Party/PartyServiceImpl.cs
```diff
@@ -2,11 +2,12 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using CSharpx;
```
jared-improbable: What's Csharpx?

services/csharp/Party/PartyServiceImpl.cs
```diff
@@ -137,23 +161,29 @@ public override async Task<JoinPartyResponse> JoinParty(JoinPartyRequest request
...
                 if (playerInvites == null)
                 {
-                    throw new RpcException(new Status(StatusCode.FailedPrecondition, "The player is not invited to this party"));
+                    throw new RpcException(new Status(StatusCode.FailedPrecondition,
+                        "The player is not invited to this party"));
                 }
-                var invited = (await Task.WhenAll(playerInvites.InboundInviteIds
+
+                var invites = (await Task.WhenAll(playerInvites.InboundInviteIds
                         .Select(invite => memClient.GetAsync<Invite>(invite))))
```
jared-improbable: You don't necessarily need to do this yet, but you could probably parallelize that `.Select` statement:
`.AsParallel().Select(...)`

services/csharp/Party/PartyServiceImpl.cs
ashleydavies: It adds a bunch of functional things including extension methods to `IEnumerable` - the one in use is for line 120, `party.GetMembers().ForEach(...)` since `party.GetMembers()` returns an `IEnumerable`

jared-improbable: Ah, I read over that. Can you please change that to just a plain `foreach` loop? The language already has a construct for iterating over a sequence, no need to pull in a third-party dependency!

PR #79, opened 2019-08-22
removed pycrypto vulnerability & tweaked endpoint spec context

LoekL:


LoekL:
I've fully tested the library change, everything is working fine. Related task: https://improbableio.atlassian.net/browse/OS-554

PR #88, opened 2019-09-18
Docs deployment pool fixes

Joe-Improbable:
- Fixed sentence in deployment pool setup guide
- Updated deployment square brackets in deployment pool doc to match the new curly braces in config file

### DIFF
--- a/docs/content/configuration-examples/deployment-pool/usage.md
+++ b/docs/content/configuration-examples/deployment-pool/usage.md
@@ -7,7 +7,7 @@ The document assumes you have already completed the [Quickstart]({{urlRoot}}/con
 
 ## Configuration
 
-The Deployment pool requires information about the deployments it is to start. These are passed as flags to the Deployment pool when it starts up. The full list of configuration parameters is as follows:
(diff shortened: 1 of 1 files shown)


docs/content/configuration-examples/deployment-pool/usage.md
```diff
@@ -7,7 +7,7 @@ The document assumes you have already completed the [Quickstart]({{urlRoot}}/con
 
 ## Configuration
 
-The Deployment pool requires information about the deployments it is to start. These are passed as flags to the Deployment pool when it starts up. The full list of configuration parameters is as follows:
+The Deployment pool requires information about the deployments it is going to start. These are passed as flags to the Deployment pool when it starts up. The full list of configuration parameters are as follows:
```
JoelOtter: This sentence was, in my opinion, grammatically correct. "List is as follows" is valid, as far as I can tell via Google, and the "is to start" is also valid, though I don't mind that one being changed if you think it reads nicer. :)

Joe-Improbable: After re-reading I agree with "List is as follows". 
For the other example I think it reads a bit nicer.

PR #89, opened 2019-09-18
fix analytics sender

LoekL:
This PR introduces three things:

- It pulls `playerId` out of eventAttributes, and turns it into an optional argument to pass in while using the `.Send()` method. If it is passed, it is now a root key of the event JSON object, instead of being inside `eventAttributes`. If it is not passed, it is simply not present in the JSON object. This change was requested by NWX.
- The analytics configuration was not working properly, disabling events did not actually prevent them from being sent. This is now fixed.
- `eventTimestamp` is upgraded from ToUnixTimeSeconds to ToUnixTimeMilliseconds, which is actually very helpful when there are loads of events being generated & you are sorting them by `eventTimestamp`.

### DIFF
--- a/services/csharp/Common/Analytics/AnalyticsSender.cs
+++ b/services/csharp/Common/Analytics/AnalyticsSender.cs
@@ -74,22 +74,23 @@ namespace Improbable.OnlineServices.Common.Analytics
         /// <summary>
         /// Queues an event, leaving any dispatching work to be done in the background rather than synchronously.
         /// </summary>
-        public void Send<T>(string eventClass, string eventType, Dictionary<string, T> eventAttributes)
+        public void Send<T>(string eventClass, string eventType, Dictionary<string, T> eventAttributes, string playerId = null)
         {
             // If the queue is to be dispatched, we leave it processing in the background instead of waiting for it
-            var _ = SendAsync(eventClass, eventType, eventAttributes);
+            var _ = SendAsync(eventClass, eventType, eventAttributes, playerId);
         }
 
         /// <summary>
         /// Queues an event for sending. If the queue was full, returns a task corresponding to sending the data to
         /// the endpoint.
         /// </summary>
-        public async Task SendAsync<T>(string eventClass, string eventType, Dictionary<string, T> eventAttributes)
+        public async Task SendAsync<T>(string eventClass, string eventType, Dictionary<string, T> eventAttributes, string playerId = null)
         {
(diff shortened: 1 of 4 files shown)


services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -101,21 +102,25 @@ public async Task SendAsync<T>(string eventClass, string eventType, Dictionary<s
         }
 
         private Dictionary<string, string> PostParams<T>(string eventClass, string eventType,
-            Dictionary<string, T> eventAttributes, long eventId)
+            Dictionary<string, T> eventAttributes, long eventId, string playerId = null)
         {
-            return new Dictionary<string, string>
+            IDictionary<string, string> eventDict = new Dictionary<string, string>
```
ashleydavies: Nit but I would `var eventDict = ...` for brevity

services/csharp/Common/Analytics/AnalyticsSender.cs
```diff
@@ -101,21 +102,25 @@ public async Task SendAsync<T>(string eventClass, string eventType, Dictionary<s
...
                 { "eventType", eventType },
-                { "sessionId", _sessionId },
                 // TODO: Add versioning ability & resolve matching TODO in relevant unit tests
                 { "versionId", "0.2.0" },
-                { "eventTimestamp", DateTimeOffset.UtcNow.ToUnixTimeSeconds().ToString() },
+                { "eventTimestamp", (DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() / (decimal) 1000).ToString("F") },
                 { "eventAttributes", JsonConvert.SerializeObject(eventAttributes) },
             };
+            if (!String.IsNullOrEmpty(playerId)) {
+                eventDict.Add(new KeyValuePair<string, string>("playerId", playerId));
```
ashleydavies: I believe `eventDict.Add("playerId", playerId)` should work, rather than manually instantiating the KeyValuePair

services/csharp/Common/Analytics/AnalyticsSender.cs
LoekL: Changed!

services/csharp/Common/Analytics/AnalyticsSender.cs
LoekL: Done :)

LoekL:
Merged this one in already so i can rebase the other PR's on `service-analytics` & properly review!

PR #90, opened 2019-09-18
instrumented deployment pool

LoekL:
This PR instruments the Deployment Pool.

Hook overview: https://docs.google.com/spreadsheets/d/1nlEWy5b97IMdpmHXWLrv5ZtuU3YnU3Ba6-JmVBxB238/edit#gid=1273567465

### DIFF
--- a/services/csharp/DeploymentPool/DeploymentPool.cs
+++ b/services/csharp/DeploymentPool/DeploymentPool.cs
@@ -8,6 +8,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Grpc.Core;
+using Improbable.OnlineServices.Common.Analytics;
 using Improbable.SpatialOS.Deployment.V1Alpha1;
 using Improbable.SpatialOS.Snapshot.V1Alpha1;
 using Serilog;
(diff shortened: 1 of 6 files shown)


services/csharp/DeploymentPool/DeploymentPool.cs
```diff
@@ -42,6 +45,7 @@ public class DeploymentPool
             cleanup = args.Cleanup;
             this.platformInvoker = platformInvoker;
             this.deploymentServiceClient = deploymentServiceClient;
+            _analytics = (analytics ?? new NullAnalyticsSender()).WithEventClass("deployment");
```
benclive: I remember something in the style guide about variables either having a `_` or not having one. I can't remember is the correct way though...

services/csharp/DeploymentPool/PlatformInvoker.cs
```diff
@@ -18,8 +19,10 @@ namespace DeploymentPool
 {
     public class PlatformInvoker
     {
+        
         private readonly DeploymentServiceClient deploymentServiceClient;
         private readonly SnapshotServiceClient snapshotServiceClient;
+        private readonly AnalyticsSenderClassWrapper _analytics;
```
benclive: same here. Pretty sure this should just be `analytics`

LoekL: I'm adhering to what is used throughout the services.

[Stackoverflow](https://stackoverflow.com/questions/549606/what-does-variable-names-beginning-with-mean/549614): _"There's no language-defined meaning - it's just a convention some people use to distinguish instance variables from local variables."_. -> we are using it because they are instance variables.

PR #91, opened 2019-09-18
instrumented Gateway

LoekL:
This PR instruments the Gateway, and includes some changes to the data model so that `partyJoinRequests` contain a `matchRequestId`, and a playerJoinRequest contains a `partyId` and a `matchRequestId` - which facilitates better tracking.

Hook overview: https://docs.google.com/spreadsheets/d/1nlEWy5b97IMdpmHXWLrv5ZtuU3YnU3Ba6-JmVBxB238/edit#gid=0 (where `eventSource` = `gateway_gateway`).

### DIFF
--- a/services/csharp/Gateway/OperationsServiceImpl.cs
+++ b/services/csharp/Gateway/OperationsServiceImpl.cs
@@ -147,7 +147,7 @@ namespace Gateway
 
                     Reporter.CancelOperationInc();
                     
-                    IDictionary<string, string> eventAttributes = new Dictionary<string, string>
+                    var eventAttributes = new Dictionary<string, string>
                     {
                         { "partyId", partyJoinRequest.Id },
                         { "matchRequestId", partyJoinRequest.MatchRequestId },
@@ -158,12 +158,12 @@ namespace Gateway
                     {
                         if (eventType == "party_match_request_cancelled")
                         {
-                            eventAttributes.Add(new KeyValuePair<string, string>("partyPhase", "Forming")); // Todo: Update currentPhase of Party with a tx
-                            _analytics.Send(eventType, (Dictionary<string, string>) eventAttributes, partyJoinRequest.Party.LeaderPlayerId);
+                            eventAttributes.Add("partyPhase", "Forming"); // Todo: Update currentPhase of Party with a tx
+                            _analytics.Send(eventType, eventAttributes, partyJoinRequest.Party.LeaderPlayerId);
                         }
                         else
                         {
-                            _analytics.Send(eventType, (Dictionary<string, string>) eventAttributes, partyJoinRequest.Party.LeaderPlayerId);
+                            _analytics.Send(eventType, eventAttributes, partyJoinRequest.Party.LeaderPlayerId);
                         }
                     }
 
--- a/services/csharp/Gateway/OperationsServiceImpl.cs
+++ b/services/csharp/Gateway/OperationsServiceImpl.cs
@@ -1,10 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Google.LongRunning;
 using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
 using Improbable.OnlineServices.Common;
+using Improbable.OnlineServices.Common.Analytics;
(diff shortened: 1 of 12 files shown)


services/csharp/Gateway/OperationsServiceImpl.cs
```diff
@@ -143,6 +146,33 @@ public override async Task<Empty> DeleteOperation(DeleteOperationRequest request
...
 
                     Reporter.CancelOperationInc();
+                    
+                    var eventAttributes = new Dictionary<string, string>
+                    {
+                        { "partyId", partyJoinRequest.Id },
+                        { "matchRequestId", partyJoinRequest.MatchRequestId },
+                        { "queueType", partyJoinRequest.Type }
+                    };
+                    string[] eventTypes = { "party_match_request_cancelled", "player_cancels_match_request" };
```
jared-improbable: Can use `var` here and below

services/csharp/Gateway/OperationsServiceImpl.cs
```diff
@@ -143,6 +146,33 @@ public override async Task<Empty> DeleteOperation(DeleteOperationRequest request
...
+                    foreach (string eventType in eventTypes)
```
jared-improbable: It feels odd to have a `foreach` here for a two-item list, one of which is already special-cased...Why not just unroll this for now?

services/csharp/Gateway/OperationsServiceImpl.cs
```diff
@@ -105,7 +108,7 @@ public override async Task<Empty> DeleteOperation(DeleteOperationRequest request
             if (!string.Equals(request.Name, playerIdentity))
             {
                 throw new RpcException(new Status(StatusCode.PermissionDenied,
-                    "Deleting another player's operation is forbidden."));
+                    $"Deleting another player's operation is forbidden: {request.Name} vs. {playerIdentity}"));
```
jared-improbable: Drive-by? Does this need to stay in?

LoekL: Was to help debug some things, but I'll revert back to original. Will leave upgrading our error messages for a potential future PR that takes everything into consideration together.

services/csharp/Gateway/OperationsServiceImpl.cs
LoekL: Like so:

```
var eventTypes = new[] { "party_match_request_cancelled", "player_cancels_match_request" };
foreach (var eventType in eventTypes)
```

Why is this better?

services/csharp/Gateway/OperationsServiceImpl.cs
LoekL: It would become: 

```
_analytics.Send("player_cancels_match_request", eventAttributes, partyJoinRequest.Party.LeaderPlayerId);
eventAttributes.Add("partyPhase", "Forming");
_analytics.Send("party_match_request_cancelled", eventAttributes, partyJoinRequest.Party.LeaderPlayerId);
```

I do not have a strong opinion on it. The foreach loop makes it very clear you're only adding `("partyPhase", "Forming")` for the party event, but it's 9 lines (vs. 3)..

jared-improbable: Using a `foreach` like is akin to building a mini state machine, and the code just doesn't warrant that.

I'd also send a copy of `eventAttributes`, or there's a world where we add a third event type, and accidentally send along unrelated data
`new Dictionary<string,string>(eventAttributes) { "partyPhase", "Forming" }`
That also makes it clearer you're adding some extra data

LoekL: Updated to:

```
_analytics.Send("player_cancels_match_request", eventAttributes, partyJoinRequest.Party.LeaderPlayerId);
var eventAttributesParty = new Dictionary<string,string>(eventAttributes) { {"partyPhase", "Forming"} };
_analytics.Send("party_match_request_cancelled", eventAttributesParty, partyJoinRequest.Party.LeaderPlayerId);
```

services/csharp/Gateway/OperationsServiceImpl.cs
LoekL: ^ No longer relevant.

PR #95, opened 2019-09-18
instrumented sample matcher

LoekL:
This PR contains the changes to instrument the sample matcher.

Hook overview: https://docs.google.com/spreadsheets/d/1nlEWy5b97IMdpmHXWLrv5ZtuU3YnU3Ba6-JmVBxB238/edit#gid=0 (where `eventSource` = `gateway_matcher`).

### DIFF
--- a/services/csharp/SampleMatcher/Matcher.cs
+++ b/services/csharp/SampleMatcher/Matcher.cs
@@ -92,7 +92,6 @@ namespace Improbable.OnlineServices.SampleMatcher
             }
             catch (RpcException e)
             {
-                Console.WriteLine(e.ToString());
                 if (e.StatusCode != StatusCode.ResourceExhausted && e.StatusCode != StatusCode.Unavailable)
                 {
                     throw;
--- a/services/csharp/SampleMatcher/Matcher.cs
+++ b/services/csharp/SampleMatcher/Matcher.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using Grpc.Core;
+using Improbable.OnlineServices.Common.Analytics;
 using Improbable.OnlineServices.Proto.Gateway;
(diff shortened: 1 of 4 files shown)


services/csharp/SampleMatcher/Matcher.cs
```diff
@@ -64,6 +92,7 @@ public Matcher()
             }
             catch (RpcException e)
             {
+                Console.WriteLine(e.ToString());
```
Zomtorg: Could be removed I guess?

services/csharp/SampleMatcher/Matcher.cs
```diff
@@ -16,11 +17,13 @@ public class Matcher : Improbable.OnlineServices.Base.Matcher.Matcher
...
         private readonly string ReadyTag = "ready"; // This should be the same tag a DeploymentPool looks for.
         private readonly string InUseTag = "in_use";
+        private static IAnalyticsSender _analytics;
 
-        public Matcher()
+        public Matcher(IAnalyticsSender analytics = null)
         {
             _project = Environment.GetEnvironmentVariable("SPATIAL_PROJECT");
             _tag = Environment.GetEnvironmentVariable("MATCH_TAG") ?? DefaultMatchTag;
+            _analytics = analytics ?? new NullAnalyticsSender();
```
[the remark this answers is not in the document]
LoekL: In this case, the `eventSource` is always the same: `gateway_matcher`. Some events are on behalf of a Party & Player(s), which have eventClass `match`, whereas the other one is on behalf of a deployment, which therefore has the eventClass set to `deployment`.

services/csharp/SampleMatcher/Matcher.cs
LoekL: Ah yes, I think that's a thing we added while debugging. Removed!

PR #97, opened 2019-09-18
updated python code

LoekL:
These changes:

- Pull `playerId` out of `eventAttributes`, which was requested by NWX.
- Add a fallback mechanism whenever the returned object is still GZIP'ed, by then first unzipping it before parsing it.


services/python/analytics-pipeline/src/dataflow/common/functions.py
```diff
@@ -2,7 +2,8 @@
 import hashlib
 import json
 import time
-
+import gzip
+import io
```
davej-improbable: Nit, but good to keep these imports ordered.

PR #99, opened 2019-09-19
Enable Nunit test reporting in buildkite

Joe-Improbable:
Adds report that shows which tests failed on buildkite

Tests

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -1,22 +1,37 @@
 #!/usr/bin/env bash
 
 # https://explainshell.com
-set -euo pipefail
+set -uo pipefail
 [[ -n "${DEBUG-}" ]] && set -x
 cd "$(dirname "$0")/../"
+
+PROJECT_DIR="$(pwd)"
+TEST_RESULTS_DIR="${PROJECT_DIR}"
 
 pushd services/csharp
 dotnet restore --no-cache
 dotnet build
-dotnet test --filter ".Test"
+TEST_FAILED=0
+for n in `find -maxdepth 1 -type d -name '*.Test'`; 
+do
+    cd ${n};
+    if ! dotnet test --logger:"nunit;LogFilePath=${TEST_RESULTS_DIR}/${n}.xml"; then
+        TEST_FAILED=1
+    fi
+    cd ../;
+done
+
+set -e
 
 pushd IntegrationTest
 export SPATIAL_PROJECT="gamex"
 export SPATIAL_REFRESH_TOKEN
 SPATIAL_REFRESH_TOKEN=$(imp-ci secrets read --environment=production --buildkite-org=improbable --secret-type=spatialos-service-account --secret-name="online-services-test" | jq -Mr '.token')
-export PLAYFAB_SECRET_KEY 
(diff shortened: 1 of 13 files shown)


services/csharp/IntegrationTest/test.sh
```diff
@@ -89,27 +89,27 @@ docker-compose -f docker_compose.yml start
...
 if [ ${test_playfab_auth} -eq 0 ]; then
   echo "Running tests for PlayFab Auth system."
-  dotnet test --filter "PlayFabAuthShould"
+  dotnet test --filter "PlayFabAuthShould" --logger:"nunit;LogFilePath=${TEST_RESULTS_DIR}/PlayFabAuth.Integration.Test.xml"
 fi
 
 if [ ${test_performance} -eq 0 ]; then
   echo "Running Performance tests."
-  dotnet test --filter "GatewayPerformanceShould"
+  dotnet test --filter "GatewayPerformanceShould" --logger:"nunit;LogFilePath=${TEST_RESULTS_DIR}/GatewayPerformance.Integration.Test.xml"
```
JoelOtter: You need to make this change in the `ps1` file too.

services/csharp/IntegrationTest/PartySystemShould.cs
```diff
@@ -69,6 +69,7 @@ public void ReturnPermissionDeniedErrorIfPitNotProvided()
             Assert.AreEqual(StatusCode.PermissionDenied, exception.StatusCode);
         }
 
+
```
JoelOtter: looks like a mistake

ci/build.sh
```diff
@@ -1,22 +1,37 @@
...
+TEST_RESULTS_DIR="${PROJECT_DIR}"
+
 pushd services/csharp
 dotnet restore --no-cache
 dotnet build
-dotnet test --filter ".Test"
+TEST_FAILED=0
+for n in `find -maxdepth 1 -type d -name '*.Test'`; 
+do
+    cd ${n};
```
JoelOtter: use `pushd`/`popd` here if you need to move between directories. If it's possible to just execute the tests with a reference to `$n` then do that?

ci/build.sh
```diff
@@ -1,22 +1,37 @@
...
+    if ! dotnet test --logger:"nunit;LogFilePath=${TEST_RESULTS_DIR}/${n}.xml"; then
+        TEST_FAILED=1
```
JoelOtter: You could actually increment this instead of setting it to 1, then we'd exit with the number of failed tests. Doesn't really matter though.

services/csharp/IntegrationTest/test.ps1
```diff
@@ -55,34 +55,39 @@ if ($rebuild) {
 	Build-Images @("gateway","gateway-internal","test-matcher","party","playfab-auth")
 }
 
+if ($null -eq $Env:TEST_RESULTS_DIR) {
```
JoelOtter: Shouldn't this be present in the `test.sh` too?

Joe-Improbable: Yep, I've added it now

PR #106, opened 2019-09-20
Fix party tests

LoekL:


LoekL:
BK stuff still failing!

services/csharp/Party/PartyServiceImpl.cs
```diff
@@ -223,12 +223,12 @@ public override async Task<JoinPartyResponse> JoinParty(JoinPartyRequest request
...
-                        {
-                            { "inviteId", invite.Id },
-                            { "playerIdInviter", invite.SenderId }
-                        })
+                     {
+                         "invites", invites.Select(invite => new Dictionary<string, string>
+                         {
+                             { "inviteId", invite.Id },
+                             { "playerIdInviter", invite.SenderId }
+                         })
```
davej-improbable: Intentional indentations?

services/csharp/Common.Test/AnalyticsShould.cs
```diff
@@ -112,8 +112,8 @@ private bool SendAnalyticEventsToHttpsEndpointExpectedMessage(HttpRequestMessage
                 Assert.True(Guid.TryParse(content.sessionId.Value, out Guid _));
 
                 // Check the timestamp is within 5 seconds of now (i.e. roughly correct)
-                long unixTimestampDelta = DateTimeOffset.UtcNow.ToUnixTimeSeconds()
-                                          - long.Parse(content.eventTimestamp.Value);
+                decimal unixTimestampDelta = (DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() / (decimal) 1000) -
+                                             decimal.Parse(content.eventTimestamp.Value);
```
davej-improbable: Why the change from `long` to `decimal`?

services/csharp/Common.Test/Common.Test.csproj
```diff
@@ -12,6 +12,7 @@
         <PackageReference Include="NUnit" Version="3.10.1" />
         <PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
         <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.1.0" />
+        <PackageReference Include="NunitXml.TestLogger" Version="2.1.41" />
```
davej-improbable: Is this required?

services/csharp/Party/PartyServiceImpl.cs
LoekL: I thought this happened when I ran `dotnet format` - but not sure. Changed back and re-ran `dotnet format`, which didn't alter it. So - I decided to stick with the original code!

services/csharp/Common.Test/AnalyticsShould.cs
LoekL: Because I changed it to decimal here as well: https://github.com/spatialos/online-services/blob/service-analytics/services/csharp/Common/Analytics/AnalyticsSender.cs#L117

Basically I wanted the highest granularity I could get, so then when I sort my events by eventTimestamp the order is correct. In case I had a ton of events (when for instance tracking the loop between the matcher & gateway-internal) I had many events on a similar truncated timestamp - making it hard to see what was actually happening.

services/csharp/Common.Test/Common.Test.csproj
LoekL: Yes, else we get this error: `Could not find a test logger with AssemblyQualifiedName, URI or FriendlyName 'nunit'.`

It's because of this line: https://github.com/spatialos/online-services/blob/service-analytics/ci/build.sh#L17 -> `--logger:"nunit;`

services/csharp/Common.Test/AnalyticsShould.cs
LoekL: (it was failing when it was still on `long`)

PR #108, opened 2019-09-23
explicitly set the default match tag used by matcher

LoekL:
Did a QA pass today re-rolling out all services. I wanted to try using a different queue type than `match` and see whether the system still worked using the SampleClient.

I updated:
- The deployment pool setup [here](https://github.com/spatialos/online-services/blob/master/services/k8s/deployment-pool/deployment.yaml#L56
).
- The SampleClient [here](https://github.com/spatialos/online-services/blob/master/services/csharp/SampleClient/Program.cs#L104).

This wasn't enough, because I forgot the Matcher is defaulting to `match` unless you pass it a `MATCH_TAG` environment variable through its k8s configuration file. This wasn't obvious because our template isn't doing this, and is instead currently relying on [this default](https://github.com/spatialos/online-services/blob/master/services/csharp/SampleMatcher/Matcher.cs#L23).

Suggest we add this value in our general configMap and capture it in the k8s config file of the sample matcher. This way it's more obvious this value also needs to be updated!

### DIFF
--- a/services/k8s/sample-matcher/deployment.yaml
+++ b/services/k8s/sample-matcher/deployment.yaml
@@ -33,5 +33,10 @@ spec:
             configMapKeyRef:
               name: online-services-config
               key: spatialos.project
+        - name: MATCH_TAG
+          valueFrom:
+            configMapKeyRef:
+              name: online-services-config
+              key: matcher.default-tag
         - name: GATEWAY_SERVICE_TARGET
           value: "gateway-internal.default.svc.cluster.local:80"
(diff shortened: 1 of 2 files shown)


services/k8s/sample-matcher/deployment.yaml
```diff
@@ -33,5 +33,10 @@ spec:
             configMapKeyRef:
               name: online-services-config
               key: spatialos.project
+        - name: MATCH_TAG
```
JoelOtter: Not sure about using a config map for default values, honestly. Weird semantics. Would just put it in directly.

LoekL: Yeah that works too!

PR #111, opened 2019-09-30
Upgrade gRPC packages across solution to 1.22.1

JoelOtter:
Upgrade `Grpc.Core` and `Grpc.Tools` across solution to 2.24.0. This is equivalent to gRPC version 1.24.0; the [major version bump](https://github.com/grpc/proposal/blob/master/L57-csharp-new-major-version.md) was caused by a breaking changed needed to support .NET Core 3.0.


JoelOtter:
Actually need to roll this back to `1.22.1`. This should fix @LoekL's issue, but there's a problem with using the 2.x.x release caused by a bug in the Platform SDK. Added details to [this ticket](https://improbableio.atlassian.net/browse/OS-651).

PR #112, opened 2019-09-30
Update Party phase to Forming when PartyJoinRequest is cancelled

Zomtorg:


services/csharp/IntegrationTest/MatchmakingSystemShould.cs
```diff
@@ -110,6 +110,49 @@ public void AllowAJoinRequestToBeDeleted()
...
+
+            // Join matchmaking.
+            var op = _gatewayClient.Join(new JoinRequest
+            {
+                MatchmakingType = "no_match"
+            }, _leaderMetadata);
+            Assert.AreEqual(LeaderPlayerId, op.Name);
+            Assert.False(op.Done);
+
+            // Verify that the party has not been matched yet. (Possibly flaky? The party might have been matched
```
JoelOtter: This won't flake; you can remove this comment. There's no matcher looking for `no_match` tags, so they'll sit in the queue til they get deleted.

services/csharp/IntegrationTest/MatchmakingSystemShould.cs
```diff
@@ -110,6 +110,49 @@ public void AllowAJoinRequestToBeDeleted()
...
+            // Verify that we can invite another member to the party
+            var pitAnotherMember = CreatePlayerIdentityTokenForPlayer(MemberPlayerId);
+            var inviteAnotherPlayer = _inviteClient.CreateInvite(new CreateInviteRequest { ReceiverPlayerId = MemberPlayerId },
+                _leaderMetadata).InviteId;
+            Assert.NotNull(inviteAnotherPlayer);
+            _partyClient.JoinParty(new JoinPartyRequest { PartyId = partyId },
+                new Metadata { { PitRequestHeaderName, pitAnotherMember } });
+
+            // Clean-up.
+            _partyClient.DeleteParty(new DeletePartyRequest(), _leaderMetadata);
```
JoelOtter: Could you also add a verification that this party can rejoin the queue?

services/csharp/IntegrationTest/MatchmakingSystemShould.cs
Zomtorg: Removed comments 👍

services/csharp/IntegrationTest/MatchmakingSystemShould.cs
Zomtorg: Added a second matchmaking join request - that should be good enough?

JoelOtter: yep!

PR #113, opened 2019-10-01
Use unique docker network per build

JoelOtter:


ci/build.sh
```diff
@@ -28,8 +28,9 @@ SPATIAL_REFRESH_TOKEN=$(imp-ci secrets read --environment=production --buildkite
 export PLAYFAB_SECRET_KEY
 PLAYFAB_SECRET_KEY=$(imp-ci secrets read --environment=production --buildkite-org=improbable --secret-type=playfab-secret-key --secret-name="online-services-playfab-secret-key-integ" | jq -Mr '.token')
 export TEST_RESULTS_DIR
+export COMPOSE_NETWORK_SUFFIX="${BUILDKITE_BUILD_ID}"
```
jared-improbable: I'd recommend using `BUILDKITE_JOB_ID` instead, since `BUILDKITE_BUILD_ID` can be reused in the case where a user presses Retry.

JoelOtter: Oh, I thought it was the other way round! Cheers.

PR #117, opened 2019-10-02
updated docker_componse.yml

LoekL:
Aligned with `master` following https://github.com/spatialos/online-services/pull/113


services/csharp/IntegrationTest/docker_compose.yml
```diff
@@ -77,16 +70,9 @@ services:
       - "D6DE8"
       - --spatial_project
       - "${SPATIAL_PROJECT}"
-networks:
-      - compose-integrationtest-network
+
```
JoelOtter: are these newlines supposed to be here?

LoekL: I'd say yes for the one on line 56, as it's also there in `master`: https://github.com/spatialos/online-services/blob/master/services/csharp/IntegrationTest/docker_compose.yml#L56

The one on 73 was a setting of my text editor..: https://github.com/atom/atom/issues/4741 - removed it!

services/csharp/IntegrationTest/docker_compose.yml
```diff
@@ -77,16 +70,9 @@ services:
...
     environment:
       SPATIAL_REFRESH_TOKEN: "${SPATIAL_REFRESH_TOKEN}"
-      GATEWAY_SERVICE_TARGET: "gateway-internal:4040"
-    networks:
-      - compose-integrationtest-network
-
-networks:
-  compose-integrationtest-network:
-    name: compose-integrationtest-network
+      GATEWAY_SERVICE_TARGET: "gateway-internal:4040"
```
LoekL: @JoelOtter - now GitHub is actually complaining: "No newline at end of file"

JoelOtter: oh we should have newlines at the end. I meant the ones you added into the middle of the file

LoekL: fixed!

PR #120, opened 2019-10-03
Service analytics

JoelOtter:


services/csharp/Gateway/GatewayServiceImpl.cs
```diff
@@ -17,10 +19,13 @@ namespace Gateway
     public class GatewayServiceImpl : GatewayService.GatewayServiceBase
     {
         private readonly IMemoryStoreClientManager<IMemoryStoreClient> _memoryStoreClientManager;
+        private static AnalyticsSenderClassWrapper _analytics;
```
JoelOtter: Why is this `static` and not `readonly`?

PR #121, opened 2019-10-04
split build, unit and integration tests into separate CI steps

Joe-Improbable:
Unit and Integration tests now run in parallel


JoelOtter:
Do the test jobs need to depend on the `build` job?

Joe-Improbable:
> Do the test jobs need to depend on the `build` job?

Nope, dotnet test will build any files it requires.

I added the build step just to make sure that any non-tested projects, would be checked to see if they were buildable

PR #124, opened 2019-10-10
Update builder queue to support Kubernetes tests

davej-improbable:


JoelOtter:
think you need to rebase?

davej-improbable:
> think you need to rebase?

Whoops yes very much so.

PR #136, opened 2019-12-09
Adopt k8s affinity fix from NWX

LoekL:


JoelOtter:
So my understanding of this is "prefer to schedule the analytics endpoint pod on a node that is not running an analytics service deployment". Is that correct?

ScottBrooks:
Yeah, the topologyKey is the "group by" basically.

When trying to find a node, group by the hostname, and try to find a node that is not currently running the node.

When using pre-emptible nodes, all of our pods would migrate to the same node, and then when that node would finally be killed we would have a bit of downtime.

LoekL:
> Yeah, the topologyKey is the "group by" basically.
> 
> When trying to find a node, group by the hostname, and try to find a node that is not currently running another version of the service.
> 
> When using pre-emptible nodes, all of our pods would migrate to the same node, and then when that node would finally be killed we would have a bit of downtime.

I found [this article](https://medium.com/@betz.mark/herding-pods-taints-tolerations-and-affinity-in-kubernetes-2279cef1f982) helpful.

In this case the topology key is `kubernetes.io/hostname` which is unique per node. So what this configuration does is saying _"do not schedule this pod on a node that already has a pod running that is labeled `app: service-analytics`"_. This helps avoid the issue @ScottBrooks just outlined.

Now you can go fancier and use some other topology key that is actually shared across several nodes, such as region. This way only one node in a particular regional node will ever run this pod, and it'll schedule the next one in a different region.

PR #137, opened 2019-12-10
updated terraform syntax in line with 0.12 to avoid warnings

LoekL:


LoekL:
> Looks good! I assume you've tried running a new plan with it :)

Yep, worked

PR #139, opened 2019-12-13
optimized cloud function to be able to use smallest template

LoekL:
This PR optimizes the analytics Cloud Function's memory usage so it can use [the smallest / cheapest template](https://cloud.google.com/functions/pricing) (128 MB / 200 MHz).

The biggest file size our Cloud Endpoint currently accepts is 32MB - I have tested and verified files of this size still succeed. Another large scale test with many files of much smaller size also succeeded.

I've changed two additional minor things:

* Switched to use the latest Python 3 string formatting structure.
* It no longer silently INFO logs BigQuery insertion errors, but raises an Exception.

### DIFF
--- a/services/python/analytics-pipeline/src/function/main.py
+++ b/services/python/analytics-pipeline/src/function/main.py
@@ -56,6 +56,7 @@ def ingest_into_native_bigquery_storage(data, context):
 
     # Parse list:
     malformed, failed_insertion = False, False
+    # We use generators in order to save memory usage, allowing the Cloud Function to use the smallest capacity template:
     for chunk in generator_chunk(generator_split(data, '\n'), 1000):
         events_batch_function, events_batch_debug = [], []
         for event_tuple in generator_load_json(chunk):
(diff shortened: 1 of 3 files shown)


services/python/analytics-pipeline/src/function/main.py
```diff
@@ -23,80 +23,90 @@ def ingest_into_native_bigquery_storage(data, context):
...
+        table_logs, table_debug, _, table_function = source_bigquery_assets(client_bq, bigquery_asset_list)
     except Exception:
-        table_logs, table_debug, table_dataflow, table_function = generate_bigquery_assets(client_bq, bigquery_asset_list)
+        table_logs, table_debug, _, table_function = generate_bigquery_assets(client_bq, bigquery_asset_list)
 
     # Parse payload:
     payload = json.loads(base64.b64decode(data['data']).decode('utf-8'))
     bucket_name, object_location = payload['bucket'], payload['name']
-    gspath = 'gs://{bucket_name}/{object_location}'.format(bucket_name=bucket_name, object_location=object_location)
+    gspath = f'gs://{bucket_name}/{object_location}'
```
JoelOtter: not seen this formatting before, nice

PR #141, opened 2019-12-17
Endpoint playfab support

LoekL:
Adding logic so that the Endpoint can handle PlayFab's event webhook forwarding system.


services/python/analytics-pipeline/src/endpoint/main.py
```diff
@@ -58,7 +58,12 @@ def store_event_in_gcs(bucket=bucket, bucket_name=os.environ['ANALYTICS_BUCKET_N
             # Parse list:
             if isinstance(payload, list):
                 for index, event in enumerate(payload):
-                    success, tried_event = try_format_event(index, event, batch_id_json, analytics_environment)
+
+                    if event_category != 'playfab':
```
JoelOtter: Random nit: I would do this the other way round, like

```python
if event_category == 'playfab':
    stuff
else:
    other_stuff
```

This way it'll be cleaner to add in additional event handling later on if we want to.

PR #148, opened 2020-01-07
Upgrade BuildKite queue, fix analytics CI

JoelOtter:
The new BuildKite images run a Goss status check on port 8080, so this also tweaks the local docker-compose configuration to expose the analytics service on host port 9090.


JoelOtter:
Just want @LoekL's all-clear to change that port!

PR #149, opened 2020-01-11
Update CI agent targeting rules

petemounce:
Added some missing definitions to more tightly target agents. https://docs.improbable.io/internal/platform/product-groups/dev-workflow/buildkite/how-to-autoscale-agents-to-match-demand
* node_stability = interruptible so that there's an 80% reduction in compute spend for the tradeoff of an agent node being shut down at random within 24 hours of being started. Since your jobs all seem to be sub five minutes, that seems like a very worthwhile tradeoff.
* make environment overridable via a var
* specify a machine_type to be explicit (same naming as spatialOS node sizes)
* specify a platform to be explicit (we're planning on soon making it possible for a single queue to contain more than one image)

Add another BK-system-error code for auto-retry.

### DIFF
--- a/.buildkite/premerge.steps.yaml
+++ b/.buildkite/premerge.steps.yaml
@@ -2,15 +2,20 @@
 common: &common
   agents:
     - "capable_of_building=online-services"
-    - "environment=production"
+    - "environment=${CI_ENVIRONMENT:-production}"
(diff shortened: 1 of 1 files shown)


petemounce:
@jared-improbable yes please.

PR #150, opened 2020-01-12
add build step

LoekL:
Added missing build step!


LoekL:
> Looks like your build failed due to an auth issue on BuildKite - maybe double-check your GitHub account is still linked correctly?

I'm using China hardware, so that's why..

PR #154, opened 2020-02-27
Add --environment flag to sampleclient

JoelOtter:


JoelOtter:
@MarkWoulfeImprobable would you be able to pull this docs change into the new ones?

PR #161, opened 2020-10-29
Update links to docs

anne-edwards:
Karl spotted that these were wrong - I've just updated them.

### DIFF
--- a/tools/ServiceAccountCLI/README.md
+++ b/tools/ServiceAccountCLI/README.md
@@ -1,3 +1,3 @@
 # The SpatialOS Online Services: Service account CLI tool
 
-See the documentation [Online Services section of the SpatialOS documentation website](https://documentation.improbable.io/online-services/docs), specifically the [Service account CLI tool documentation](https://documentation.improbable.io/online-services/docs/platform-service-account-cli).
+See the documentation [Online Services section of the SpatialOS documentation website](https://online-services.readme.io/docs/welcome), specifically the [Service account CLI tool documentation](https://online-services.readme.io/docs/platform-service-account-cli).


tools/ServiceAccountCLI/README.md
```diff
@@ -1,3 +1,3 @@
 # The SpatialOS Online Services: Service account CLI tool
 
-See the documentation [Online Services section of the SpatialOS documentation website](https://documentation.improbable.io/online-services/docs), specifically the [Service account CLI tool documentation](https://documentation.improbable.io/online-services/docs/platform-service-account-cli).
+See the documentation [Online Services section of the SpatialOS documentation website](https://online-services.readme.io/docs/welcome), specifically the [Service account CLI tool documentation](https://online-services.readme.io/docs/platform-service-account-cli).
```
tomtkarl, line 3: These links present me with a login page. That right?

tomtkarl:
Lots of broken docs links in this repo, fwiw: https://github.com/spatialos/online-services/search?q=online-services&type=code

tools/ServiceAccountCLI/README.md
anne-edwards: Hmm, good point. I think the plan was to give access to anyone who needed it, via a ReadMe account. I'll give you access now - or at least try to (with ReadMe it's never as easy as it should be).

tomtkarl: Wait, so we have to invite users to create a readme account to read these docs? Can these docs not be public, given that the tools it is documenting are public on Github?

anne-edwards: That's another good point. I'm not 100% sure of the background to this move, but from what you're saying, it sounds like we can keep these docs public but just not have them on our main docs site. Will double-check with the other TWs first thing tomorrow in case I've missed some context, but it sounds like it should be fine.

JoelOtter: I'm confused. Are docs not public any more?

JoelOtter: Also...why does that not redirect?

JoelOtter:
Hi folks, I had a chat to Nathan and we're going to deprecate & archive this repo now - we don't think anyone's using it and it doesn't really have an owner. @anne-edwards could you remove its documentation from the Readme side if that's something that needs done over there?

anne-edwards:
@JoelOtter @tomtkarl now I'm slightly confused too! It looks like Karl was using this repo (and presumably some others are too?) and also using the Online Services docs. We (tech writers) were planning to make the docs public (just not visible from our main docs site) and update all the broken links. Do we still need to do this?

(Background: currently, the docs are not visible from our main docs site, but they're also private, which means we need to invite people to make a ReadMe account in order to see them. I _think_ this might have been an oversight back when we moved them out of the main set of docs.)

JoelOtter:
Ah, gotcha. Let me get back to you on this one. :)

